PR #13, opened 2021-02-15
Add Bash Build Script

colton-smith:
## Brief
Added a simple bash script to build and run the sample application.


liamjwarren (approved):
Very nice! Just tested this on MacOS along with PR 14 changes and it works. This will make it nice and easy for submissions. I will update README.md for a section for Windows, Mac, and Linux.

Burke-Daniel (approved):
Worked like a charm when I tested it. GTM

PR #14, opened 2021-02-15
Fix error with macro expansion on GCC

Burke-Daniel:
Fixes error @colton-smith was seeing on linux, basically MSVC was smart and when `__VA_ARGS__` is used with no arguments supplied, it omits the would be trailing comma from the macro expansion.

For example, `AQ_WARNING("Could not create window");` expands to `::Aquarius::Log::getClientLogger()->warn("Could not create window")` on windows but `::Aquarius::Log::getClientLogger()->warn("Could not create window", )` on linux(gcc). The fix to this is to include` ##` in front of `__VA_ARGS__`, which omits the comma if no arguments are supplied (MSVC preprocessor seems to do this automatically).

Also removed semicolons from macro definitions even though it didn't seem to be breaking anything

### DIFF
--- a/Aquarius/Src/Aquarius/Core/Log.h
+++ b/Aquarius/Src/Aquarius/Core/Log.h
@@ -5,18 +5,18 @@
 // Note: for printf-like formatting, use only %v instead of type specific identifiers within log string
 
 // Core log macros
-#define AQ_CORE_TRACE(String, ...)      ::Aquarius::Log::getCoreLogger()->trace(String, __VA_ARGS__);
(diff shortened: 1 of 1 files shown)


Burke-Daniel:
@liamjwarren wanna make sure this works on MacOS as well?

colton-smith (approved):
Tested on WSL, old error gone but cant say for sure there aren't any new ones.

GTM I'd say

liamjwarren (approved):
All good on MacOS, GTM

PR #15, opened 2021-02-23
Assertion wrapper

Burke-Daniel:
This PR wraps assertions up with some variadic template magic. The assertions will only be compiled if `AQ_ASSERT_ENABLE` is defined, otherwise these are just functions with empty bodies (in other words, nothing).

### DIFF
--- a/Aquarius/Aquarius.cpp
+++ b/Aquarius/Aquarius.cpp
@@ -1,3 +1,5 @@
+#define AQ_ASSERT_ENABLE
+
 #include "Aquarius.h"
 #include "Aquarius/Core/Log.h"
 #include "Aquarius/Core/Window.h"
--- a/Aquarius/Aquarius.cpp
+++ b/Aquarius/Aquarius.cpp
@@ -3,6 +3,8 @@
 #include "Aquarius.h"
 #include "Aquarius/Core/Log.h"
 #include "Aquarius/Core/Window.h"
+#include "Aquarius/Core/Assert.h"
+
 
 namespace Aquarius {
 
--- a/Aquarius/Src/Aquarius/Core/Assert.h
+++ b/Aquarius/Src/Aquarius/Core/Assert.h
@@ -12,6 +12,7 @@ inline void aqCoreAssert(bool expression)
 	if (!expression)
 	{
 		AQ_CORE_ERROR("Assertion failed at %v:%v", __FILE__, __LINE__);
+		std::abort();
 	}
 	#endif
 }
@@ -24,6 +25,7 @@ inline void aqCoreAssert(bool expression, Msg message, Ts ... args)
 	{
 		std::string errorMessage = "Assertion failed: " + std::string(message);
 		AQ_ERROR(errorMessage.c_str(), args...);
+		std::abort();
 	}
 	#endif
 }
@@ -34,6 +36,7 @@ inline void aqAssert(bool expression)
 	if (!expression)
 	{
 		AQ_ERROR("Assertion failed at %v:%v", __FILE__, __LINE__);
+		std::abort();
 	}
 	#endif
 }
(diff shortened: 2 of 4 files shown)


Aquarius/Aquarius.cpp
```diff
@@ -30,6 +33,12 @@ int Test::testMain()
     AQ_INFO("Testing %v %v", "client", "logger");
     AQ_CORE_INFO("Testing %v %v", "core", "logger");
 
+    aqAssert(5 == 4);
+    aqAssert(4 == 5, "This is obviously %v", "wrong");
+
+    aqCoreAssert(true == false, "This is obviously false");
+    aqCoreAssert("hello" == "world", "This is obviously %v", "wrong");
+
```
colton-smith: After testing, execution is not stopped or paused after this assertion, can we somehow do that?

Aquarius/Aquarius.cpp
```diff
@@ -1,5 +1,8 @@
+#define AQ_ASSERT_ENABLE
```
colton-smith, line 1: Eventually these kind of definitions will be set based on the build mode (debug/release). Thats some house keeping we'll have to do later on

Aquarius/Src/Aquarius/Core/Assert.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+#include "Aquarius/Core/Log.h"
+
+#include <string>
+
+namespace Aquarius {
+
+inline void aqCoreAssert(bool expression)
```
colton-smith, line 9: NGL I dont know what an inline function is

colton-smith (changes requested):
Currently, this seems very similar to just typing 
```
if (!expression)
{
AQ_ERROR(some msg_
}

```

Is there any way we can make this stop / pause execution like `assert` from assert.h?

Aquarius/Aquarius.cpp
Burke-Daniel: Yeah that is something that can occur within CMake

Aquarius/Src/Aquarius/Core/Assert.h
Burke-Daniel: Basically just think of it in terms of the assembly code that gets output; the compiler can either create a function that the assembly can jump to and reuse, or if it feels it is appropriate to do so, when a function is declared inline it will just put that function call directly "inline" with the assembly

Burke-Daniel: Good for stuff like this, small trivial functions. But just because you declare it as inline doesn't mean it will be inlined, it is just a suggestion for the compiler

Aquarius/Aquarius.cpp
Burke-Daniel: Yeah so now `std::abort()` is used, which is what the assert macro uses to terminate program execution

Burke-Daniel:
Back to you for another look @colton-smith, after that will wait to merge until other PRs get merged first

liamjwarren (approved):

Aquarius/Src/Aquarius/Core/Assert.h
```diff
@@ -0,0 +1,56 @@
...
+}
+
+template<typename Msg, typename ... Ts>
+inline void aqCoreAssert(bool expression, Msg message, Ts ... args)
+{
+	#ifdef AQ_ASSERT_ENABLE
+	if (!expression)
+	{
+		std::string errorMessage = "Assertion failed: " + std::string(message);
+		AQ_ERROR(errorMessage.c_str(), args...);
```
colton-smith: Should call AQ_CORE_ERROR

colton-smith (approved):
See comment then GTM

PR #16, opened 2021-02-23
Windowing and Rendering Context

colton-smith:
This PR adds an initial (mostly complete) implementation of a `Window` class and a `RenderingContext` class. Their lifetime are tied, by way of Window having a unique pointer to a rendering context. Other than readability / making it easier to abstract down the road, these could have been just one class. 

Importantly:
- `Window` handles window creation 
- `RenderingContext` loads OpenGL symbols and sets current context to the created window
- `Window` destructor destroys the glfw window and terminates glfw
- Vsync can be enabled/disabled 

TODO
- Register callbacks

### DIFF
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -86,6 +86,9 @@ namespace Aquarius {
     void Window::Deallocate()
     {
         glfwDestroyWindow(m_Window);
+        glfwTerminate();
+
+        AQ_CORE_INFO("Window destroyed and GLFW terminated");
     }
 
     void Window::setVsync(bool enabled)
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -44,9 +44,8 @@ namespace Aquarius {
         m_Window = glfwCreateWindow(m_Width, m_Height, m_Name.c_str(), NULL, NULL);
         if (m_Window == nullptr)
         {
-            AQ_CORE_FATAL("Failed to create a glfw window");
             glfwTerminate();
-            assert(false);
+            AQ_CORE_FATAL("Failed to create a glfw window");
         }
 
         // Generate context
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -15,6 +15,13 @@ namespace Aquarius {
 
     void Window::Initialize()
     {
+        // Set GLFW error callback
+        glfwSetErrorCallback([](int error_code, const char* description)
+                             {
+                                 AQ_CORE_ERROR("GLFW Error Occurred:");
+                                 AQ_CORE_ERROR(description);
+                             });
+
         // Initialize GLFW
         int initSuccess = glfwInit();
         if (!initSuccess)
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -5,7 +5,12 @@
 namespace Aquarius {
 
     Window::Window(uint32_t width, uint32_t height, std::string &&name, bool vsync)
-        : m_Width(width), m_Height(height), m_Name(std::move(name)), m_VsyncEnabled(vsync), m_Window(nullptr), m_Context(nullptr)
+        : m_Width(width),
+          m_Height(height),
+          m_Name(std::move(name)),
+          m_VsyncEnabled(vsync),
+          m_Window(nullptr),
+          m_Context(nullptr)
     {}
 
     Window::WindowPtr Window::Create(uint32_t width, uint32_t height, std::string &&name, bool vsync)
@@ -106,7 +111,7 @@ namespace Aquarius {
         }
         else
         {
-            glfwSwapInterval(1);
+            glfwSwapInterval(0);
             AQ_CORE_INFO("Vsync disabled");
         }
         m_VsyncEnabled = enabled;
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -73,7 +73,7 @@ namespace Aquarius {
         return m_Name;
     }
 
-    bool Window::getVsync() const
+    bool Window::isVsyncEnabled() const
     {
         return m_VsyncEnabled;
     }
--- a/Aquarius/Src/Aquarius/Core/Window.cpp
+++ b/Aquarius/Src/Aquarius/Core/Window.cpp
@@ -1,0 +1,106 @@
+#include "Window.h"
+#include "Log.h"
+#include <assert.h>
+
+namespace Aquarius {
+
+    Window::Window(uint32_t width, uint32_t height, std::string &&name, bool vsync)
+        : m_Width(width), m_Height(height), m_Name(std::move(name)), m_VsyncEnabled(vsync), m_Window(nullptr), m_Context(nullptr)
+    {}
+
+    Window::WindowPtr Window::Create(uint32_t width, uint32_t height, std::string &&name, bool vsync)
+    {
+        return std::make_unique<Window>(width, height, std::move(name));
+    }
+
+    void Window::Initialize()
+    {
+        // Initialize GLFW
+        int initSuccess = glfwInit();
+        if (!initSuccess)
+        {
+            AQ_CORE_FATAL("Failed to initialize GLFW");
+        }
+        else
+        {
+            AQ_CORE_INFO("GLFW successfully initialized");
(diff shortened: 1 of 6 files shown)


Aquarius/Src/Aquarius/Renderer/RenderingContext.cpp
```diff
@@ -0,0 +1,45 @@
...
+            });
+        }
+        else
+        {
+            AQ_CORE_INFO("GLAD successfully initialized");
+        }
+
+        AQ_CORE_INFO("Vendor: %v", glGetString(GL_VENDOR));
+        AQ_CORE_INFO("GPU: %v", glGetString(GL_RENDERER));
+        AQ_CORE_INFO("OpenGL Version: %v", glGetString(GL_VERSION));
```
Burke-Daniel, line 31: Would just like to say that this is super sick to see getting logged haha

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
+#include "Window.h"
+#include "Log.h"
+#include <assert.h>
+
+namespace Aquarius {
+
+    Window::Window(uint32_t width, uint32_t height, std::string &&name, bool vsync)
+        : m_Width(width), m_Height(height), m_Name(std::move(name)), m_VsyncEnabled(vsync), m_Window(nullptr), m_Context(nullptr)
```
Burke-Daniel: What are we going with for max line length? Not that it matters but could put each of these on their own line

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {}
+
+    Window::WindowPtr Window::Create(uint32_t width, uint32_t height, std::string &&name, bool vsync)
```
Burke-Daniel, line 16: I like wrapping the std::unique_ptr here, looks a lot cleaner

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
...
+        return std::make_unique<Window>(width, height, std::move(name));
+    }
+
+    void Window::Initialize()
+    {
+        // Initialize GLFW
+        int initSuccess = glfwInit();
+        if (!initSuccess)
+        {
+            AQ_CORE_FATAL("Failed to initialize GLFW");
```
Burke-Daniel, line 34: This is good because this will cause the program to terminate

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
...
+        #ifdef PLATFORM_MAC
+                glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE);
+        #endif // PLATFORM_MAC
+
+        m_Window = glfwCreateWindow(m_Width, m_Height, m_Name.c_str(), NULL, NULL);
+        if (m_Window == nullptr)
+        {
+            AQ_CORE_FATAL("Failed to create a glfw window");
+            glfwTerminate();
+            assert(false);
```
Burke-Daniel: Might want to double check, but I believe AQ_CORE_FATAL will cause program termination, in that case, may wanna remove the assert and move the call to `glfwTerminate()` to above the log that was all the opengl stuff exits cleanly before terminate

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return m_Height;
+    }
+
+    const std::string& Window::getName() const
+    {
+        return m_Name;
+    }
+
+    bool Window::getVsync() const
```
Burke-Daniel: I know its a bit more verbose, but would something like `isVsyncEnabled` be more representative of what this does?

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (enabled)
+        {
+            glfwSwapInterval(1);
+            AQ_CORE_INFO("Vsync enabled");
+        }
+        else
+        {
+            glfwSwapInterval(1);
+            AQ_CORE_INFO("Vsync disabled");
+        }
```
Burke-Daniel: So without knowing what `glfwSwapInterval` does, it seems like it just toggles vsync. Could just move that call to above the conditional block since both calls are the exact same

Aquarius/Src/Aquarius/Core/Window.h
```diff
@@ -0,0 +1,46 @@
...
+
+    // GLFW Window
+    class Window
+    {
+        using WindowPtr = std::unique_ptr<Window>;
+
+    public:
+        // Creation
+        Window(uint32_t width, uint32_t height, std::string&& name, bool vsync = false);
+        static WindowPtr Create(uint32_t width, uint32_t height, std::string&& name, bool vsync = false);
```
Burke-Daniel, line 17: Default args here for vsync, nice!

Aquarius/Src/Aquarius/Core/Window.h
```diff
@@ -0,0 +1,46 @@
...
+
+        // Utility methods
+        uint32_t getWidth() const;
+        uint32_t getHeight() const;
+        const std::string& getName() const;
+        bool getVsync() const;
+        void setVsync(bool enabled);
+
+        // Destruction
+        ~Window();
```
Burke-Daniel: I like how you have ordered the methods almost "chronologically" in a way, makes it super easy to read

Aquarius/Src/Aquarius/Renderer/RenderingContext.cpp
```diff
@@ -0,0 +1,45 @@
...
+        if (!gladLoadGLLoader((GLADloadproc)glfwGetProcAddress))
+        {
+            AQ_CORE_FATAL("Failed to initialize GLAD" );
+
+            // Set GLFW error callback
+            glfwSetErrorCallback([](int error_code, const char* description)
+            {
+                AQ_CORE_ERROR("GLFW Error Occurred:");
+                AQ_CORE_ERROR(description);
+            });
```
Burke-Daniel: Switch the order of these so that the error callback code actually gets reached before program termination.
P.S. nice lambda

Burke-Daniel (changes requested):
I had more comments then necessary, just being super nit picky and admiring the lambdas and pretty logging lol, but do see my comment about `AQ_FATAL` and `AQ_CORE_FATAL` causing program termination

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: I was thinking about that yea, I dont think we should worry too much about 80 chars but just keep it reasonable hahaha

Burke-Daniel: Agreed, and with so many arguments I think this may look stupid when they are all on their own line so it can probably stay as-is

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: Yeah std::unique_ptr<RenderingContext>and the like is pretty long and annoying

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: Didn't realize fatal would terminate until after this, good point

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: Yeah thats very UE4, I will do that

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: Jesus that was supposed to be glfwSwapInterval(0)

Aquarius/Src/Aquarius/Core/Window.cpp
liamjwarren: Is this line too long? It goes past the standard length line on my IDE.

liamjwarren (commented):
Just one style comment, not sure if it something we are concerned about. Daniel did a pretty good job of covering the major things.

Aquarius/Src/Aquarius/Core/Window.h
colton-smith: This was intentional

Aquarius/Src/Aquarius/Core/Window.cpp
colton-smith: Reduced the length, it was pretty gross before yeah

liamjwarren (approved):

Burke-Daniel (approved):
GTM

PR #17, opened 2021-02-23
Logging refinement

Burke-Daniel:
This PR adds the ability to set the logging level, the logging levels are defined as follows (from least to greatest severity):
```
Trace
Info
Warning
Error
Fatal
```
Note: Fatal logs cause the program to terminate

The ability to set a logging output file for both the core and client loggers is now possible, with the default being that the loggers do not log their output to a file, but just the console.

P.S. base branch is set to assertion-wrapper just so those changes don't show up in the diff (I rebased on top of it)

### DIFF
--- a/Aquarius/Aquarius.cpp
+++ b/Aquarius/Aquarius.cpp
@@ -2,10 +2,7 @@
 
 #include "Aquarius.h"
 #include "Aquarius/Core/Log.h"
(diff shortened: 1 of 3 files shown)


liamjwarren (approved):

colton-smith (approved):
GTM

PR #18, opened 2021-02-24
Initial Implementation of Event System

### LINKED ISSUE #3: Design and implement event system

liamjwarren:
Initial Commit:
Completed
-Initial Keyboard event callbacks implemented.
-Initial Mouse event callbacks implemented.
-Logging completed for all callbacks, can be tested by building and information will appear for all callbacks.
-Set callbacks in Aquarius.cpp for Keyboard and Mouse completed.

TODO (in order)
-Implement Window event initial callbacks
-Set callbacks in Aquarius.cpp for Window.
-Expand event system
-Update logging messages to include key and mouse button names rather than their int representation.


Aquarius/Src/Aquarius/Events/Keyboard.h
```diff
@@ -0,0 +1,32 @@
...
+#include <GLFW/glfw3.h>
+#include "Aquarius/Core/Log.h"
+
+namespace Aquarius {
+    class KeyboardEvent {
+    public:
+        static void Callback(GLFWwindow* window, int key, int scanCode, int action, int modifications)
+        {
+            //TODO: Get it so key button name is printed rather then number, this is not critical but would be
+            // nice to have.
```
Burke-Daniel, line 14: Not sure if you thought through how to accomplish this or if you have seen anyone else do it, but a `static const std::unordered_map<int, std::string>` comes to mind that you could throw in here as a lookup table

Aquarius/Aquarius.cpp
```diff
@@ -27,6 +30,16 @@ int Test::testMain()
...
 
+    //TODO Bring in neccesary set methods for window events.
+    glfwSetKeyCallback(window, KeyboardEvent::Callback);
+    glfwSetCursorPosCallback(window, MouseEvent::cursorPositionCallback);
+    glfwSetCursorEnterCallback(window, MouseEvent::cursorInWindowCallback);
+    glfwSetMouseButtonCallback(window, MouseEvent::mouseButtonCallback);
+    glfwSetScrollCallback(window, MouseEvent::mouseScrollCallback);
+
+    glfwSetInputMode(window, GLFW_STICKY_MOUSE_BUTTONS, 1);
+    glfwSetInputMode(window, GLFW_STICKY_KEYS,1);
```
Burke-Daniel: All this should eventually happen within some of our internal code, either in the application or window initialization I would say

Aquarius/Src/Aquarius/Events/Keyboard.h
```diff
@@ -0,0 +1,32 @@
...
+                case GLFW_PRESS:
+                    AQ_INFO("key %v is pressed.", key);
+                    break;
+
+                case GLFW_RELEASE:
+                    AQ_INFO("key %v has been released.", key);
+                    break;
+
+                case GLFW_REPEAT:
+                    AQ_INFO("key %v is being pressed repeatedly. ", key);
```
Burke-Daniel: Extra space here at the end of the string

Aquarius/Src/Aquarius/Events/Mouse.h
```diff
@@ -0,0 +1,55 @@
...
+        {
+            // TODO: Discuss if we want to only keep track of cursor position when the cursor is on
+            //  the Application window, if so can combine this callback method and the one below.
+            AQ_INFO("x position: %v", xPosition);
+            AQ_INFO("y position: %v", yPosition);
+        }
+
+        static void cursorInWindowCallback(GLFWwindow* window, int inWindow)
+        {
+            if (inWindow)
```
Burke-Daniel, line 21: Seeing an integer interpreted as a boolean is slightly misleading, but this is a consequence of glfw being written in C (C doesn't have a bool type), so this is fine here

Aquarius/Src/Aquarius/Events/Mouse.h
```diff
@@ -0,0 +1,55 @@
...
+                AQ_INFO("Cursor has left application window.");
+            }
+        }
+
+        static void mouseButtonCallback(GLFWwindow* window, int button, int action, int modifications)
+        {
+            switch (action)
+            {
+                //TODO: Get it so mouse button name is printed rather then number, this is not critical but would be
+                // nice to have.
```
Burke-Daniel, line 37: See comment above about unordered_map

Burke-Daniel (approved):
Works as expected when I run it, see lots of logging based off of keypresses, releases, and holds, and moving my mouse around in the window. I know Colton mentioned the same line braces so I didn't comment on those.

Have you thought about how this is going to fit in to the event system? right now these classes are basically just wrapping the callbacks we have to set in glfw. In terms of passing this information around the system, it might be good to have a base `Event` abstract class that all these inherit from. We can talk about this on the call. Other than that this is GTM after those couple of small style things

Aquarius/Aquarius.cpp
liamjwarren: Good call, we can discuss the best spot for this.

PR #19, opened 2021-02-24
Core application class and entry point abstraction

Burke-Daniel:
This PR adds a base Application class that we can build off of. It also changes the Sandbox example to utilize this new Application class.

To use, a derived class must be created with a constructor that calls the base constructor (passing a std::string for the window name). As well, a createApplication function which returns a `std::unique_ptr<Application>` must be created, returning a derived unique_ptr invoking the constructor.


liamjwarren (approved):
GTM

colton-smith (approved):
GTM

PR #20, opened 2021-02-24
Add Vertex/Index Buffer Abstractions

colton-smith:
## Brief
This PR adds simple abstraction around OpenGL vertex and index buffer.

## Summary
- `VertexBuffer`: Create, bind, destroy vertex buffer given a pointer to some data 
- `IndexBuffer`: Create, bind, destroy vertex buffer given a pointer to a uint32_t array of indices

## Note
I cant test / render any thing until I add shader abstraction, so I may actually do that before I do VertexArray abstraction.


Burke-Daniel (approved):
As talked about on call, GTM

liamjwarren (approved):
GTM

PR #21, opened 2021-02-25
Fix size_t error on Clang

liamjwarren:
-Added #include <stddef.h> into IndexBuffer.h and VertexBuffer.h to resolve size_t error being returned by clang.


Burke-Daniel (approved):
GTM

PR #22, opened 2021-02-25
Add Input Polling

Burke-Daniel:
Separate from input events, this PR adds input polling to the engine. It allows for keys, mouse buttons, and the mouse cursor position to be polled for at any time. The Application class was updated to terminate the main application loop when the escape key is pressed, and the Window class was updated to include a get() method for grabbing a pointer to the GLFWwindow.


Aquarius/Src/Aquarius/Core/Input.cpp
```diff
@@ -0,0 +1,39 @@
...
+	KeyState getKeyState(KeyCode key)
+	{
+		auto window = Application::get()->getWindow();
+		return static_cast<KeyState>(glfwGetKey(window->get(), static_cast<int>(key)));
+	}
+
+	MouseButtonState getMouseButtonState(MouseButtonCode button)
+	{
+		auto window = Application::get()->getWindow();
+		return static_cast<MouseButtonState>(glfwGetMouseButton(window->get(), static_cast<int>(button)));
```
colton-smith: This won't implicitly cast because you are using enum class and not enum right?

colton-smith (approved):
Looks good dude! The only thing I was wondering was if we wanted an `Input` class / namespace to contain this stuff? (class would be static only)

`Aquarius::getKeyState` vs `Aquarius::Input::getKeyState` is the only difference. Doesn't matter to me, just something to consider

Aquarius/Src/Aquarius/Core/Input.cpp
Burke-Daniel: Correct yeah

Burke-Daniel:
Yeah I'll go ahead and move this into an Input namespace so that we don't pollute our Aquarius namespace too much

Aquarius/Src/Aquarius/Core/Input.h
```diff
@@ -0,0 +1,168 @@
...
+namespace Aquarius {
+	enum class KeyState : int
+	{
+		Released = 0,
+		Pressed = 1
+	};
+
+	using MouseButtonState = KeyState;
+
+	enum class KeyCode : int
```
liamjwarren: Looks good, once you merge I will update my refinement branch so I can use these in the event files.

Aquarius/Src/Aquarius/Core/Input.h
```diff
@@ -0,0 +1,168 @@
...
+		Mouse_button_left = Mouse_button_1,
+		Mouse_button_right = Mouse_button_2,
+		Mouse_button_middle = Mouse_button_3
+	};
+
+	struct MousePosition
+	{
+		double x;
+		double y;
+	};
```
liamjwarren: Once we get to dealing with mouse scrolling inputs we will need to make a similar struct. Not pertinent to the additions you made but I am just thinking ahead.

Aquarius/Src/Aquarius/Core/Input.h
```diff
@@ -0,0 +1,168 @@
...
+	enum class KeyState : int
+	{
+		Released = 0,
+		Pressed = 1
+	};
+
+	using MouseButtonState = KeyState;
+
+	enum class KeyCode : int
+	{
```
liamjwarren: Make a mention that these are all from glfw3.h

Aquarius/Src/Aquarius/Core/Input.h
```diff
@@ -0,0 +1,168 @@
...
+	};
+
+	KeyState getKeyState(KeyCode key);
+	MouseButtonState getMouseButtonState(MouseButtonCode button);
+	bool isKeyPressed(KeyCode key);
+	bool isMouseButtonPressed(MouseButtonCode button);
+	MousePosition getMousePosition();
+} // namespace Aquarius
+
+
```
liamjwarren, line 164: Additional newlines at EOF

liamjwarren (approved):
Looks good, just a few minor comments.

Aquarius/Src/Aquarius/Core/Input.h
Burke-Daniel: Good point

Aquarius/Src/Aquarius/Core/Input.h
Burke-Daniel: Yup good call

PR #23, opened 2021-02-25
Add Shader Abstraction 

colton-smith:
## Brief
This PR adds a class that abstracts shader compilation and linking. It updates the test code to use the shader abstraction. I was able to test the vertex buffer and index buffer classes using this and everything worked well (although the test code isn't executable anywhere in the repo right now, when layers are added I can move the test to the SandboxLayer and we will have a nice place to start developing a sample application)

## Summary
- Provided a link to a vertex and shader source file, can compile and link them
- Compile/link errors are logged
- File IO errors are caught and logged
- Following compiling and linking, can bind, unbind, and destroy the shader program


Aquarius/Src/Aquarius/Renderer/Shader.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        if (!linkSuccess)
+        {
+            char linkLog[512];
+            glGetProgramInfoLog(m_ID, 512, NULL, linkLog);
+            AQ_CORE_ERROR("Error linking shader");
+            AQ_CORE_ERROR("Linking error: %v", linkLog);
+        }
+    }
+} // namespace Aquarius
```
liamjwarren: Add newline at EOF

Aquarius/Src/Aquarius/Renderer/Shader.h
```diff
@@ -0,0 +1,53 @@
...
+        };
+
+    private:
+        void compile(uint32_t shaderID, const char* shaderCode, ShaderType shaderType);
+        void link(uint32_t vertexID, uint32_t fragmentID);
+        uint32_t m_ID;
+    };
+} // namespace Aquarius
+
+
```
liamjwarren: Unnecessary additional newlines at EOF

Aquarius/Aquarius.cpp
```diff
@@ -62,22 +68,23 @@ namespace Aquarius {
         glBindVertexArray(VA0);
 
         // Create and bind vertex buffer
-        VertexBuffer VB0 = VertexBuffer(data.data(), data.size() * sizeof(data[0]));
+        VertexBuffer VB0 = VertexBuffer(data.data(), data.size() * sizeof(float));
```
Burke-Daniel, line 71: This is very minor, but if you're gonna invoke the constructor on the RHS of the expression like that, i'm partial to using `auto VB0` instead, cause the type is still very evident based off of the variable name and the constructor

Aquarius/Src/Aquarius/Renderer/Shader.h
```diff
@@ -0,0 +1,53 @@
...
+        {
+            VERTEX_SHADER,
+            FRAGMENT_SHADER
+        };
+
+        std::unordered_map<ShaderType, std::string> shaderTypeToString =
+        {
+            {ShaderType::VERTEX_SHADER, "VERTEX_SHADER"},
+            {ShaderType::FRAGMENT_SHADER, "FRAGMENT_SHADER"}
+        };
```
Burke-Daniel: Could probably be `static const`

Aquarius/Src/Aquarius/Renderer/Shader.h
```diff
@@ -0,0 +1,53 @@
...
+#include <unordered_map>
+
+namespace Aquarius {
+
+    class Shader
+    {
+    public:
+        // Path to vertex and fragment shader source code
+        Shader(const std::string& vertexPath, const std::string& fragmentPath);
+        ~Shader();
```
Burke-Daniel, line 17: Should we maybe delete the default constructor? I think the compiler will generate one where the only data member is a uint32

Aquarius/Src/Aquarius/Renderer/Shader.cpp
```diff
@@ -0,0 +1,177 @@
...
+    {
+        int location = glGetUniformLocation(m_ID, name.c_str());
+        glUniform4i(location, values.x, values.y, values.z, values.w);
+    }
+
+    void Shader::setMat4(const std::string& name, bool transpose, const glm::mat4& value) const
+    {
+        int location = glGetUniformLocation(m_ID, name.c_str());
+        glUniformMatrix4fv(location, 1, transpose, glm::value_ptr(value));
+    }
```
Burke-Daniel, line 141: If only glfw decided to use C++/templates :(

Aquarius/Src/Aquarius/Renderer/Shader.cpp
Burke-Daniel: EOF

Burke-Daniel (approved):
Couple of small comments but then GTM!

liamjwarren (approved):

Aquarius/Src/Aquarius/Renderer/Shader.cpp
liamjwarren: Ahhh great minds think alike...

Aquarius/Aquarius.cpp
colton-smith: I will definitely do that, sick point

Aquarius/Src/Aquarius/Renderer/Shader.h
colton-smith: Yes very much so, I was thinking static for sure but const didnt even cross my mind

colton-smith: When I implemented this, I learned a lot! Init'ing the static member outside the header, then not being able to use the [] operator because it isn't const, I had to use .at()!

Aquarius/Src/Aquarius/Renderer/Shader.h
colton-smith: I tested this, it doesnt create a constructor with uint32 parameter or even a default one with no params (super weird). I think this can for sure happen sometimes though, according to the vague docs ("Sometimes a constructor is implicitly declared")

I will delete the default constructor tho, probably good practice

Aquarius/Src/Aquarius/Renderer/Shader.cpp
colton-smith: Hahaha

PR #24, opened 2021-02-25
Cleanup and Add Utilities

colton-smith:
This is a PR that looks like a big diff but it's not, it just touches a lot of files with spacing changes. I added a similar wrapping of unique and shared pointer as The Chernos Ref and Scope, without the fancy wrapping of std::make_unique and shared with variatic arguments (I'm thinking we can just use std::make_unique when we have to).

Apart from that, I updated the includes with new spacing and ordering, and discovered some unsused includes I could remove. I also added a definition to the Aquarius and Sandbox cmakelists to avoid dumb `glad already included` errors (and tested) .

Finally, I included all of the engine headers in Aquarius.h so clients can just include one file and get all of the engine stuff (usually this would be a precompiled header).

If anything I did makes no sense, LMK! :) 

P.S. 
Instead of looking at the diff in one file view, this one might be better viewed as a side by side diff due to all the whitespace changes.

### DIFF
--- a/Aquarius/Aquarius.h
+++ b/Aquarius/Aquarius.h
@@ -1,11 +1,3 @@
-
-#include <iostream>
-#include <glad/glad.h>
-#include <GLFW/glfw3.h>
-#include <easylogging++.h>
-#include <stb_image.h>
-#include <glm/glm.hpp>
-
 // -- Core
 #include "Aquarius/Core/Utility.h"
 #include "Aquarius/Core/Application.h"
@@ -22,6 +14,14 @@
 #include "Aquarius/Events/Keyboard.h"
 #include "Aquarius/Events/Mouse.h"
 #include "Aquarius/Events/Window.h"
+
+#include <iostream>
+#include <glad/glad.h>
+#include <GLFW/glfw3.h>
+#include <easylogging++.h>
+#include <stb_image.h>
+#include <glm/glm.hpp>
+
 
 namespace Aquarius {
     class Test
--- a/Aquarius/Aquarius.h
+++ b/Aquarius/Aquarius.h
@@ -6,6 +6,22 @@
 #include <stb_image.h>
 #include <glm/glm.hpp>
 
+// -- Core
+#include "Aquarius/Core/Utility.h"
+#include "Aquarius/Core/Application.h"
+#include "Aquarius/Core/Assert.h"
+#include "Aquarius/Core/Input.h"
+#include "Aquarius/Core/Log.h"
+#include "Aquarius/Core/Window.h"
+
+// -- Renderer
+#include "Aquarius/Renderer/IndexBuffer.h"
+#include "Aquarius/Renderer/VertexBuffer.h"
+
+// -- Events
+#include "Aquarius/Events/Keyboard.h"
+#include "Aquarius/Events/Mouse.h"
+#include "Aquarius/Events/Window.h"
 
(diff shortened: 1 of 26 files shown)


Aquarius/Aquarius.h
```diff
@@ -6,6 +6,22 @@
 #include <stb_image.h>
 #include <glm/glm.hpp>
 
+// -- Core
```
liamjwarren, line 1: Nice and clean, I like it.

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -1,10 +1,10 @@
 #include "Window.h"
 #include "Log.h"
-#include <assert.h>
 #include "Aquarius/Events/Keyboard.h"
 #include "Aquarius/Events/Mouse.h"
 #include "Aquarius/Core/Input.h"
```
liamjwarren: Shouldn't the core stuff be above all this?

Aquarius/Src/Aquarius/Core/Application.cpp
```diff
@@ -1,7 +1,7 @@
-
 #include "Application.h"
-#include "Aquarius/Core/Log.h"
 #include "Aquarius/Core/Input.h"
```
liamjwarren, line 2: If the file is in the same sub-directory as the includes are we including the path or just the name? We have done this differently throughout.

Aquarius/Src/Aquarius/Core/Application.h
```diff
@@ -1,9 +1,11 @@
 #pragma once
 
 #include "Aquarius/Core/Window.h"
```
liamjwarren: See comment about include standards for the same subdirectory.

Aquarius/Src/Aquarius/Core/Window.h
```diff
@@ -1,20 +1,21 @@
 #pragma once
+
+#include "Aquarius/Renderer/RenderingContext.h"
+#include "Aquarius/Core/Utility.h"
```
liamjwarren, line 3: See comment about include standards for the same subdirectory.

Aquarius/Src/Aquarius/Core/Window.h
```diff
@@ -1,20 +1,21 @@
 #pragma once
+
+#include "Aquarius/Renderer/RenderingContext.h"
```
liamjwarren: This should be below core

liamjwarren (changes requested):
Good idea to have set some standards and cleaned up the code to date as we get back to school next week. I just have a few minor changes and one we can discuss if necessary is the way we want to do our includes from the same subdirectory. Right now in some places, we do the full path and in others, we just use the name of the file.

Aquarius/Src/Aquarius/Core/Window.cpp
liamjwarren: A good example here of what I mean by the two styles of includes for the same subdirectory here.

colton-smith: @liamjwarren Just wondering why you say that, because its local or because its alphabetical? I think we are gonna go with alphabetical for this

liamjwarren (approved):

Burke-Daniel (approved):
Looks all GTM

PR #25, opened 2021-02-26
Vertex Array Object Abstraction

colton-smith:
This PR adds abstraction of the OpenGL vertex array object, using 3 classes. The vertex array object stores a state which includes a bound vertex buffer, index buffer, and most importantly vertex buffer layout which tells the GPU how to interpret the data in the vertex buffer.

You can create a VAO, add a vertex and index buffer, then set the layout (there is also the option to construct the VAO with the buffers and the layout).

I did some interesting things to get the GLenum type, creating a map between types I defined and types defined in GLenum so that the client doesn't need glad to create a buffer layout.

### VAO
- Encapsulates creating VAO, adding index/vertex buffers, and setting the vertex buffer layout

### BufferLayout
- Defines the Vertex attribute layout for a certain vertex buffer, contains a vector of VertexElements. On construction will calculate the offset of each element and the stride

### VertexElement
- Defines a single vertex element with the type of elements, the number of elements, and if the elements should be normalized.

There's some interesting things in here, would love some feedback!

### DIFF
--- a/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
@@ -5,7 +5,6 @@
 
 namespace Aquarius {
 
-    // TODO - What style to use here
     VertexArray::VertexArray(const std::shared_ptr<VertexBuffer>& vertexBuffer,
                              const std::shared_ptr<IndexBuffer>& indexBuffer,
                              const BufferLayout& bufferLayout)
@@ -13,8 +12,7 @@ namespace Aquarius {
     {
         glGenVertexArrays(1, &m_ID);
         setIndexBuffer(indexBuffer);
-        setVertexBuffer(vertexBuffer);
-        setBufferLayout(bufferLayout);
+        setVertexBuffer(vertexBuffer, bufferLayout);
     }
 
     VertexArray::VertexArray()
@@ -37,32 +35,40 @@ namespace Aquarius {
         glBindVertexArray(0);
     }
 
+    void VertexArray::setVertexBuffer(const std::shared_ptr<VertexBuffer>& vertexBuffer, const BufferLayout& bufferLayout)
+    {
+        activate();
+        vertexBuffer->Bind();
+        m_VertexBuffer = vertexBuffer;
+        setBufferLayout(bufferLayout);
+    }
+
     void VertexArray::setVertexBuffer(const std::shared_ptr<VertexBuffer>& vertexBuffer)
     {
-        m_VertexBuffer = vertexBuffer;
         activate();
-        m_VertexBuffer->Bind();
+        vertexBuffer->Bind();
+        m_VertexBuffer = vertexBuffer;
     }
 
     void VertexArray::setIndexBuffer(const std::shared_ptr<IndexBuffer>& indexBuffer)
     {
-        m_IndexBuffer = indexBuffer;
         activate();
-        m_IndexBuffer->Bind();
+        indexBuffer->Bind();
+        m_IndexBuffer = indexBuffer;
     }
 
     void VertexArray::setBufferLayout(const BufferLayout& bufferLayout)
     {
         activate();
 
-        size_t vertexAttribIndex = 0;
+        uint32_t vertexAttribIndex = 0;
         for (auto elem : bufferLayout.getElements())
         {
             glEnableVertexAttribArray(vertexAttribIndex);
             glVertexAttribPointer(vertexAttribIndex,
                                   elem.getCount(),
                                   elem.getType(),
-                                  elem.getNormalize(),
+                                  elem.getNormalize() ? GL_TRUE : GL_FALSE,
                                   bufferLayout.getStride(),
                                   (void*)elem.getOffset());
             vertexAttribIndex++;
--- a/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
@@ -62,7 +62,7 @@ namespace Aquarius {
         activate();
 
         uint32_t vertexAttribIndex = 0;
-        for (auto elem : bufferLayout.getElements())
+        for (const auto& elem : bufferLayout.getElements())
         {
             glEnableVertexAttribArray(vertexAttribIndex);
             glVertexAttribPointer(vertexAttribIndex,
--- a/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
@@ -1,0 +1,14 @@
+#include "VertexArray.h"
+
+namespace Aquarius {
+
+    VertexArray::VertexArray()
+    {
+
+    }
+
+    VertexArray::~VertexArray()
+    {
+
+    }
+} // namespace Aquarius
--- a/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
@@ -5,8 +5,8 @@
 
 namespace Aquarius {
 
-    VertexArray::VertexArray(const std::shared_ptr<VertexBuffer>& vertexBuffer,
-                             const std::shared_ptr<IndexBuffer>& indexBuffer,
+    VertexArray::VertexArray(std::shared_ptr<VertexBuffer> vertexBuffer,
+                             std::shared_ptr<IndexBuffer> indexBuffer,
                              const BufferLayout& bufferLayout)
             : m_VertexBuffer(vertexBuffer), m_IndexBuffer(indexBuffer)
     {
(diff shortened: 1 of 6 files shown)


Aquarius/Src/Aquarius/Renderer/VertexArray.h
```diff
@@ -0,0 +1,120 @@
...
+        {ShaderType::Double, GL_DOUBLE}
+    };
+
+    static const std::unordered_map<ShaderType, uint32_t>  ShaderTypeSizeBytes =
+    {
+            {ShaderType::Int, 4},
+            {ShaderType::UInt, 4},
+            {ShaderType::Float, 4},
+            {ShaderType::Double, 8}
+    };
```
Burke-Daniel: Idea to reduce a bit of repetition here, what if you had a struct called ShaderInfo or something, containing a GLenum and a uint32_t. and then that way you could have one `unordered_map<ShaderType, ShaderInfo>`? Something like:
```C++
    struct ShaderInfo
    {
        GLenum glType;
        uint32_t size;
    };

    static const std::unordered_map<ShaderType, ShaderInfo> ShaderTypeToGLEnum =
    {
        {ShaderType::Int, {GL_INT, 4}},
        {ShaderType::UInt, {GL_UNSIGNED_INT, 4}},
        {ShaderType::Float, {GL_FLOAT, 4}},
        {ShaderType::Double, {GL_DOUBLE, 8}}
    };
```

Aquarius/Src/Aquarius/Renderer/VertexArray.h
```diff
@@ -0,0 +1,120 @@
...
+        uint32_t m_Offset;
+        bool m_Normalize;
+    };
+
+    class BufferLayout
+    {
+    public:
+        BufferLayout() = delete;
+        ~BufferLayout() = default;
+        BufferLayout(std::initializer_list<VertexElement> vertexElements)
```
Burke-Daniel: Haven't seen a good ol' initializer list since last summer lmao

Aquarius/Src/Aquarius/Renderer/VertexArray.h
```diff
@@ -0,0 +1,120 @@
...
+        {
+            // Calculate offset for each element
+            uint32_t Offset = 0;
+            for (auto elem : vertexElements)
```
Burke-Daniel, line 76: I would probably make this `const auto& elem : elements` where these are objects with a good few fields. There is a chance that auto will iterate over and copy by value here. Cool thing is the more info you add when using auto, the less work the compiler has to do to figure out what you want. Either way this will prevent you from copying the elements twice

Burke-Daniel: Actually now looking at it you could potentially use move semantics here where you are passing in an initializer list of values, but idt it matters too much either way

Aquarius/Src/Aquarius/Renderer/VertexArray.cpp
```diff
@@ -0,0 +1,88 @@
...
+        indexBuffer->Bind();
+        m_IndexBuffer = indexBuffer;
+    }
+
+    void VertexArray::setBufferLayout(const BufferLayout& bufferLayout)
+    {
+        activate();
+
+        uint32_t vertexAttribIndex = 0;
+        for (auto elem : bufferLayout.getElements())
```
Burke-Daniel: Same thing, could also do `const auto& elem` here

Aquarius/Src/Aquarius/Renderer/VertexArray.h
```diff
@@ -0,0 +1,120 @@
...
+        std::vector<VertexElement> m_Elements;
+        uint32_t m_Stride;
+    };
+
+    class VertexArray
+    {
+    public:
+        VertexArray();
+        VertexArray(const std::shared_ptr<VertexBuffer>& vertexBuffer,
+                    const std::shared_ptr<IndexBuffer>& IndexBuffer,
```
Burke-Daniel: So I remember reading an article about when you should pass by value, reference, passing pointers (shared and unique), and wanted to look up what the conventional guidelines for this are. So obviously this constructor (plus the functions setVertexBuffer and setIndexBuffer) wants to obtain partial ownership of the underlying object (i.e increment the refcount), and here's what the [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-smartptrparam) say (idk if you have read these before but they are great btw):
_"R.34: Take a shared_ptr<widget> parameter to express that a function is part owner"_
_"R.35: Take a shared_ptr<widget>& parameter to express that a function might reseat the shared pointer"_
_"R.36: Take a const shared_ptr<widget>& parameter to express that it might retain a reference count to the object ???"_
So obviously R.35 doesn't apply, because we are not reassigning the shared_ptr to another object, but I think R.34 represents our situation better than R.36 because we are always retaining a refcount when these functions/constructors are called

Burke-Daniel: TLDR: I think we might wanna change these to pass by value

Burke-Daniel: BTW this helped me better understand memory ownership, I find it very interesting haha

Aquarius/Src/Aquarius/Renderer/VertexArray.h
```diff
@@ -0,0 +1,120 @@
...
+        void activate();
+        void deactivate();
+
+        void setVertexBuffer(const std::shared_ptr<VertexBuffer>& vertexBuffer);
+        void setVertexBuffer(const std::shared_ptr<VertexBuffer>& vertexBuffer, const BufferLayout& bufferLayout);
+        void setIndexBuffer(const std::shared_ptr<IndexBuffer>& IndexBuffer);
+        void setBufferLayout(const BufferLayout& bufferLayout);
+
+        std::shared_ptr<VertexBuffer> getVertexBuffer() const;
+        std::shared_ptr<IndexBuffer> getIndexBuffer() const;
```
Burke-Daniel: In the same vein, do we want to be definitely incrementing the refcount here when someone wants to view/modify the vertex/index buffers? We may wanna just return a raw pointer here

Aquarius/Src/Aquarius/Renderer/VertexArray.h
Burke-Daniel: Also missing our sharedPtr wrapper :(

Burke-Daniel (changes requested):
This is one of those PRs where I learned a lot reading it. This is super sick and lmk what you think of the couple of changes I requested!

Aquarius/Src/Aquarius/Renderer/VertexArray.h
colton-smith: Whats nutty is I didnt look at the code you suggested and then wrote it VERBATIM

Aquarius/Src/Aquarius/Renderer/VertexArray.h
colton-smith: THey are god tier

Aquarius/Src/Aquarius/Renderer/VertexArray.h
colton-smith: Where I call setOffset which is not const, it wont allow me to use `const auto& elem` or even `auto& elem`

colton-smith: Seems `auto& elem` gets interpreted as `const auto& elem`

colton-smith: https://en.cppreference.com/w/cpp/utility/initializer_list

You cant get a non const reference iterator over the list thats why

Aquarius/Src/Aquarius/Renderer/VertexArray.h
colton-smith: If this were multi threaded, it would be dangerous to use a raw pointer here I think because the ref count could hit 0 while something else is using the raw pointer obtained from these methods. In our case, everything occurs sequentially so I think we may be good with raw? Unless the renderer for some reason has to wait until the next frame to render or something. Its hard to know what to do until the implementation details come around

Burke-Daniel: Yeah that's a fair point, well in that case i'd say its ok to leave the way it is, but your call. Idt we will get around to threading anyways

colton-smith: Yeah we definitely wont, just wondering what the negative to increasing the ref count is? I guess if something just wants a peak at the VB/IB it doesnt need/want to be part owner

Burke-Daniel: Yeah exactly, so there's no real harm other than a slight performance hit (obviously the refcount has to be checked then when that returned shared_ptr goes out of scope to determine whether or not to deallocate the memory, and it may only be in scope for a very short period of time)

colton-smith: raw pointer it is

Burke-Daniel (approved):
Yep looks GTM now, gonna rebase on top of this and use it in my layering branch

PR #31, opened 2021-02-26
Add Barebones Renderer Implementation 

colton-smith:
Something to build off of, adds a static only renderer class with a submit method for drawing a provided VAO to the screen. Altered the test code in the testTriangle method as well. I plan on extending this renderer to include calls for drawing quads using screen space coordinates (pixel values between screen width and height) instead of using NDC all the time. 

@Burke-Daniel You could probably rebase on this and use the renderer calls inside SandboxLayer.

### DIFF
--- a/Aquarius/Src/Aquarius/Renderer/Renderer.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/Renderer.cpp
@@ -33,4 +33,4 @@ namespace Aquarius {
         glDrawElements(GL_TRIANGLES, vertexArray->getIndexBuffer()->getCount(), GL_UNSIGNED_INT, (void*)0);
     }
 
-}
+} // namespace Aquarius 
--- a/Aquarius/Src/Aquarius/Renderer/Renderer.cpp
+++ b/Aquarius/Src/Aquarius/Renderer/Renderer.cpp
@@ -1,0 +1,36 @@
+#include "Renderer.h"
+#include "Aquarius/Core/Application.h"
+
+#include <glad/glad.h>
+
+
+namespace Aquarius {
+
+    void Renderer::Init()
+    {
+        glPolygonMode(GL_FRONT, GL_FILL);
+    }
+
+    void Renderer::Shutdown()
+    {
+        // Nothing to put here just yet
+    }
+
+    void Renderer::Clear()
+    {
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    }
+
+    void Renderer::ClearColor(glm::vec3 color)
+    {
+        glClearColor(color.x, color.y, color.z, 1.0f);
+    }
+
+    void Renderer::Submit(VertexArray* vertexArray, Shader* shader)
+    {
+        shader->activate();
+        vertexArray->activate();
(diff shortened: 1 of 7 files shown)


Aquarius/Src/Aquarius/Renderer/Renderer.h
```diff
@@ -0,0 +1,27 @@
...
+#include <glm/glm.hpp>
+
+
+namespace Aquarius {
+
+    class Renderer
+    {
+    public:
+        Renderer() = delete;
+        ~Renderer() = delete;
```
colton-smith, line 16: The intention of this class is static only, so I deleted the default constructor. However, Im not sure if there are still other implicit constructors (copy, move, etc). 

Also, I feel like this is a terrible thing to do?

Aquarius/Src/Aquarius/Renderer/Renderer.cpp
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    void Renderer::Submit(VertexArray* vertexArray, Shader* shader)
+    {
+        shader->activate();
+        vertexArray->activate();
+        glDrawElements(GL_TRIANGLES, vertexArray->getIndexBuffer()->getCount(), GL_UNSIGNED_INT, (void*)0);
+    }
+
+}
```
liamjwarren: Add in `//  namespace Aquarius`

Aquarius/Src/Aquarius/Renderer/Renderer.h
```diff
@@ -0,0 +1,27 @@
...
+        static void Init();
+        static void Shutdown();
+
+        static void Clear();
+        static void ClearColor(glm::vec3 color);
+
+        static void Submit(VertexArray* vertexArray, Shader* shader);
+    };
+
+}
```
liamjwarren: Add in `//  namespace Aquarius`

liamjwarren (approved):
Looks good! The only thing I noticed was that `// namespace Aquarius` was missing in both Renderer files but otherwise GTM.

Aquarius/Src/Aquarius/Renderer/Renderer.h
```diff
@@ -0,0 +1,27 @@
...
+        ~Renderer() = default;
+
+        static void Init();
+        static void Shutdown();
+
+        static void Clear();
+        static void ClearColor(glm::vec3 color);
+
+        static void Submit(VertexArray* vertexArray, Shader* shader);
+    };
```
Burke-Daniel: So will objects of this class ever need to be instantiated? If so, wdyt about moving these methods into a `Renderer` namespace instead? Seems like there is no particular gain to these being in a class vs namespace as the default constructor is deleted, there are no data members, and every method is static

Burke-Daniel: Though if the plan is to extend this and add functionality that would require us instantiating the object then disregard the above haha

Burke-Daniel (approved):
One question/comment then GTM whatever you decide to do

Aquarius/Src/Aquarius/Renderer/Renderer.h
colton-smith: Yeah there's a concept called instanced rendering, although I doubt we will ever implement that. This renderer class will have static members only which can also just be defined in a namespace, so that sounds good.

PR #32, opened 2021-02-27
Initial Layering System Implementation

Burke-Daniel:
This will make our lives a lot easier! Includes a base Layer class which must be inherited from, and this will allow us to propagate events through the system, and render in a sequential order which we decide. For now, the LayerStack is just an alias to a `std::deque<uniquePtr<Layer>>`, and the application base class now has the ability to push layers onto the stack. We can add functionality as necessary. I have also pulled in @colton-smiths renderer implementation and demonstrated that in a Sandbox layer. I went ahead and removed Aquarius.cpp, Idt we will need it for testing from now on.


Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include "Aquarius/Core/Utility.h"
```
liamjwarren, line 3: Add newline between pragma and include.

Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <deque>
+#include <string>
+
```
liamjwarren, line 7: Add newline here so that there are two empty lines between last include and namespace.

Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
...
+		virtual void onCreation() {};
+		virtual void onEvent(Event& event) {};
+		virtual void updateAndRender(timeDelta_t time) {};
+		virtual void onDestruction() {};
+
+		virtual const std::string& getLayerName() const = 0;
+	};
+
+}
```
liamjwarren: Add `// namespace Aquarius`

Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <deque>
+#include <string>
+
+namespace Aquarius {
+
+	class Event;
+	class Layer;
+
+	using LayerStack = std::deque<uniquePtr<Layer>>;
```
liamjwarren, line 14: Should we call this LayerDeque instead in the essence of being 100% transparent lol

liamjwarren: Stack does sound better than Deque though lol

Aquarius/Src/Aquarius/Core/Application.h
```diff
@@ -25,9 +25,11 @@ namespace Aquarius {
 		void run();
 	protected:
 		Application(std::string&& windowName = "Application");
+		void PushLayer(uniquePtr<Layer> layer);
 	private:
 		static Application* s_Application;
 		uniquePtr<Window> m_Window;
+		LayerStack m_layerStack;
```
liamjwarren, line 33: See my comment about stack vs deque.

Aquarius/Src/Aquarius/Core/Application.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 
+#include "Aquarius/Core/Layer.h"
 #include "Aquarius/Core/Utility.h"
 #include "Aquarius/Core/Window.h"
 
 #include <memory>
```
liamjwarren: Add newline here so that there are two empty lines between last include and namespace.

Sandbox/Sandbox.h
```diff
@@ -2,6 +2,41 @@
 
 #include <Aquarius.h>
 
+#include <array>
```
liamjwarren, line 5: I know we are using this for testing right now but thinking ahead when we want people to be able to include the single Aquarius header file should we put `#include<array>`in `Aquarius.h`? That way people do not need to include it in their Sandbox as I am thinking arrays will be used in pretty much any situation as you have done below. It is not a big deal but I am just thinking ahead.

liamjwarren (approved):
Looks good man! Just a few comments regarding style standards and some thoughts I had looking ahead.

Aquarius/Src/Aquarius/Core/Application.cpp
```diff
@@ -25,14 +27,26 @@ namespace Aquarius {
 
 	void Application::run()
 	{
+		auto time = std::chrono::high_resolution_clock::now();
 		while (!glfwWindowShouldClose(m_Window->get()))
 		{
-			glClearColor(0.2f, 0.3f, 0.3f, 1.0f);
-			glClear(GL_COLOR_BUFFER_BIT);
+			auto timeDelta = std::chrono::high_resolution_clock::now() - time;
+			time = std::chrono::high_resolution_clock::now();
```
colton-smith: Should you maybe call now() once and store it in a variable? I know one line is like nanoseconds but now will be return something different in each of these calls which is logically not desired idt

Aquarius/Src/Aquarius/Core/Application.cpp
```diff
@@ -25,14 +27,26 @@ namespace Aquarius {
...
+			for (const auto& layer : m_layerStack)
+			{
+				layer->updateAndRender(std::chrono::duration_cast<std::chrono::microseconds>(timeDelta).count() / 1000.0);
+			}
+			time = std::chrono::high_resolution_clock::now();
```
colton-smith: Why overwrite the time here as well? This seems like  it will result in false FPS, timing only the delta between here and the start of the loop when really you want to time the layer iteration and update most importantly

Aquarius/Src/Aquarius/Core/Application.cpp
```diff
@@ -25,14 +27,26 @@ namespace Aquarius {
...
 			m_Window->OnUpdate();
 		}
 	}
 
+	void Application::PushLayer(uniquePtr<Layer> layer)
+	{
```
colton-smith, line 45: Wondering if you could also take a && instead of forcing the client to construct an owning ptr only to move it immediately. Would that pose any dangers / negatives?

`PushLayer(std::make_unique<SandboxLayer>(X, X, X))` vs `PushLayer(SandboxLayer(X, X, X))`

Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
...
+
+	class Layer
+	{
+	public:
+
+		virtual ~Layer() { onDestruction(); };
+
+		virtual void onCreation() {};
+		virtual void onEvent(Event& event) {};
+		virtual void updateAndRender(timeDelta_t time) {};
```
colton-smith: Partial to `onUpdate` for this call for consistency, looks a little out of place

Sandbox/Sandbox.h
colton-smith: I think it's okay to ask clients to include things from std. This is just a convenience thing, they could also have used a regular array for the vertices.

Sandbox/Sandbox.h
```diff
@@ -2,6 +2,41 @@
...
+
+class SandboxLayer : public Aquarius::Layer
+{
+public:
+	SandboxLayer(const std::string& vertexShader, const std::string& fragmentShader);
+
+	void onCreation() override;
+	void updateAndRender(Aquarius::timeDelta_t time) override;
+
+	const std::string& getLayerName() const override { return "Sandbox"; }
```
colton-smith: Can you return a const & to a temp? I would have thought this is generally a no-go

Aquarius/Src/Aquarius/Core/Layer.h
```diff
@@ -0,0 +1,29 @@
...
+	class Event;
+	class Layer;
+
+	using LayerStack = std::deque<uniquePtr<Layer>>;
+	using timeDelta_t = float;
+
+	class Layer
+	{
+	public:
+
```
colton-smith: Why not have the layer name as a field? It seems weird to have a pure virtual method that will be overridden just to return a string.

colton-smith: In the future, also considering having an enum for layer types, that way we can easily filter and order layers easily

Aquarius/Src/Aquarius/Core/Layer.h
colton-smith, line 14: Apparently deque doesn't guarantee contiguous storage and is therefore ~50% slower to iterate over than a vector. Although, our bottleneck likely isn't the iteration over the layers but rather actually calling the update method of each layer. Also, we probably won't have enough layers for this to affect us

colton-smith (approved):
Pumped to see this get in there, just a couple of comments regarding delta time (thanks for adding that btw, gonna need it very soon) and the getName method.

colton-smith (changes requested):
Meant to request changes, see previous comments (facepalm)

Aquarius/Src/Aquarius/Core/Layer.h
Burke-Daniel: I'm gonna leave it as Stack because we are only ever gonna push Layers onto one end, and pop them from that same end

Burke-Daniel: Yeah we can just leave this for now and if it causes issues then switch it to a vector down the road

Aquarius/Src/Aquarius/Core/Application.cpp
Burke-Daniel: The issue is that when I do it this way, I have to do `push_back<Layer>(std::make_unique<Layer>(std::move(layer))) which actually creates a base layer object using the implicit copy constructor, and then we lose the ability to call the overridden methods, as we actually now have a base Layer object pointed to by a Layer pointer, instead of a SandboxLayer pointed to by a Layer pointer

Aquarius/Src/Aquarius/Core/Layer.h
Burke-Daniel: Originally had it this way, was trying to avoid having data members in the base class but for a single string, probably not necessary

Burke-Daniel:
Just pushed some commits addressing comments, left a couple of comments unresolved there for you to take a look at. I'll get you to look again before I merge

colton-smith (approved):
GTM

PR #33, opened 2021-02-28
Event System Framework

### LINKED ISSUE #38: Complete the event system implementation

liamjwarren:
The event system framework is now put in place. I still need to connect it with the callback functions and the input polling to ensure it functions as intended. I also want to make a function to set all callbacks that are currently done in Window.cpp to be a function in the Callbacks.h file. I will refine further in the coming days but wanted to post to get some comments to date and see if there are any glaring issues that I need to address.


Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+        }
+    };
+
+    class MouseEvent
+    {
+    public:
+        static void cursorPositionCallback(GLFWwindow* window, double xPosition, double yPosition)
+        {
+            // TODO: Discuss if we want to only keep track of cursor position when the cursor is on
+            //  the Application window, if so can combine this callback method and the one below.
```
colton-smith: Definitely dont care if the cursor is outside the window I'd say

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,31 @@
...
+        Event() = default;
+        Event(T Type) : type(Type) {};
+        virtual ~Event() = default;
+
+        inline const T getEventType() const
+        {
+            return type;
+        }
+
+        virtual bool handledStatus()
```
colton-smith: Not sure this needs to be virtual, should be the same function definition for all sub classes

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -1,9 +1,8 @@
 #include "Window.h"
+#include "Log.h"
+#include <assert.h>
+#include "Aquarius/Events/Callbacks.h"
```
colton-smith: Should be:
-  Aquarius/Core/Log.h
-  assert.h is incorrectly placed and shouldn't be needed anyway

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+                    AQ_INFO("mouse button %v is pressed.", button);
+                    break;
+
+                case GLFW_RELEASE:
+                    AQ_INFO("mouse button %v has been released.", button);
+                    break;
+
+                default:
+                    break;
+            }
```
colton-smith: I guess these callbacks will need to emit an event of appropriate type right? Either directly use a singleton instance of EventManager or pass the event to Application where it can then call EventManager->Notify

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    class WindowEvent
+    {
+    public:
+
+        static void windowCloseCallback(GLFWwindow* window)
+        {
+            //TODO Implement
+        }
```
colton-smith: I'd say we can cut these down to what is really necessary for us

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Aquarius {
+
+    template <typename T>
+    class EventHandler
+    {
+    private:
+        using functionType = std::function<void(const Event<T>&)>;
+        std::map<T, std::vector<functionType>> subscribers;
+
```
colton-smith: When you had this after public, the public functions likely had no idea what functionType was. One option is to just move `using functionType ... ` above the public stuff

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <GLFW/glfw3.h>
+
+
+namespace Aquarius {
+
+    template <typename T>
```
colton-smith: Might need someone to explain to me why this class needs to be templated

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,45 @@
...
+
+
+namespace Aquarius {
+
+    enum class MouseEvents
+    {
+        MouseButtonPressedEvent, MouseButtonReleasedEvent, MouseMovedEvent, MouseScrolledEvent, MouseInWindowEvent
+    };
+
+    class MouseButtonPressed : public Event<MouseEvents>
```
colton-smith: Wondering if MouseButtonPressedEvent (suffix Event) might be a better way to name these events

Aquarius/Src/Aquarius/Events/Event.h
colton-smith: Oh, i think its because the event types are not all in the same enum

Burke-Daniel:
Haven't actually reviewed this fully yet, but saw a chance to use macros to reduce code repetition when defining the derived event classes for each event type. See this [branch](https://github.com/Burke-Daniel/Aquarius/tree/refine-initial-event-system-review) for what I did, and if you think it's reasonable then you can just cherry-pick the commit, if not then disregard lol

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Aquarius {
+
+    template <typename T>
+    class EventHandler
+    {
+    private:
+        using functionType = std::function<void(const Event<T>&)>;
+        std::map<T, std::vector<functionType>> subscribers;
```
colton-smith: We should also consider using unordered_map, it's much more performant and I **think** will serve the same purpose

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+        void subscribe(T Type, const functionType& function)
+        {
+            subscribers[Type].push_back(function);
+        }
+
+        void notify(Event<T>& event)
+        {
+            // Check to see if at end of subscribers
+            if((subscribers.find(event.getEventType())) == subscribers.end())
+                return;
```
colton-smith: This checks if there are any subscribers of a given event type, end of subscribers was a little misleading.

Aquarius/Src/Aquarius/Events/Event.h
colton-smith: also isHandled might be a little clearer of a name

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+                    AQ_INFO("key %v is being pressed repeatedly.", key);
+                    break;
+
+                default:
+                    break;
+            }
+        }
+    };
+
+    class MouseEvent
```
colton-smith: I think we should update these names to be MouseCallback or something. I do like defining them here and using them in the window class to register the callbacks

colton-smith (changes requested):
Very sick my guy! I think I understand how it will work, just need to plug it in in a few places and we're g2g with events!

Aquarius/Src/Aquarius/Events/Callbacks.h
liamjwarren: Yes, something along that line.

Aquarius/Src/Aquarius/Events/Event.h
liamjwarren: Correct

Aquarius/Src/Aquarius/Events/Callbacks.h
Burke-Daniel: If EventManager is a member of Application, we can use the static Application get() method to accomplish that I would say. Just one way of going about it

Aquarius/Src/Aquarius/Events/Callbacks.h
Burke-Daniel: Yeah and on that note, do each of these need to be within a class or can they just be free functions?

Aquarius/Src/Aquarius/Events/Callbacks.h
Burke-Daniel: ^^

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+    {
+    public:
+        static void cursorPositionCallback(GLFWwindow* window, double xPosition, double yPosition)
+        {
+            // TODO: Discuss if we want to only keep track of cursor position when the cursor is on
+            //  the Application window, if so can combine this callback method and the one below.
+            AQ_INFO("x position: %v", xPosition);
+            AQ_INFO("y position: %v", yPosition);
```
Burke-Daniel: Do you wanna drop the log level of this to TRACE? These will definitely just be used for debugging as they pollute the logs haha

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,135 @@
...
+
+                default:
+                    break;
+            }
+        }
+
+        static void mouseScrollCallback(GLFWwindow* window, double xOffSet, double yOffset)
+        {
+            AQ_INFO("x offset: %v", xOffSet);
+            AQ_INFO("y offset: %v", yOffset);
```
Burke-Daniel: Could drop these to TRACE as well

Burke-Daniel: Also I think any logging in here we will probably want to be CORE

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+            for(auto&& subscriber : subscribers.at(event.getEventType()))
+            {
+                // Notify of event only if the event has not already been handled
+                if(event.handledStatus() == 0)
```
Burke-Daniel: Make comparison to bool instead of int

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+    class EventHandler
+    {
+    private:
+        using functionType = std::function<void(const Event<T>&)>;
+        std::map<T, std::vector<functionType>> subscribers;
+
+    public:
+        void subscribe(T Type, const functionType& function)
+        {
+            subscribers[Type].push_back(function);
```
Burke-Daniel: For subscribers where the event will change the state of the object, we will wanna make sure to use std::bind, and bind member functions and a respective instantiated object here

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,33 @@
...
+    {
+    public:
+        KeyRepeated() : Event<KeyboardEvents>(KeyboardEvents::KeyRepeatedEvent) {};
+    };
+
+    class KeyReleased : public Event<KeyboardEvents>
+    {
+    public:
+        KeyReleased() : Event<KeyboardEvents>(KeyboardEvents::KeyReleasedEvent) {};
+    };
```
Burke-Daniel: As discussed, will have to add data about the type of key pressed

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,45 @@
...
+    {
+    public:
+        MouseScrolled() : Event<MouseEvents>(MouseEvents::MouseScrolledEvent) {};
+    };
+
+    class MouseInWindow : public Event<MouseEvents>
+    {
+    public:
+        MouseInWindow() : Event<MouseEvents>(MouseEvents::MouseInWindowEvent) {};
+    };
```
Burke-Daniel: Same as above, will have to add info about what button is pressed

Burke-Daniel (changes requested):
A lot of these changes we already discussed today, so once those are changed I'd like to have another look. Also, seems like the commit history got a bit messy there. No big deal but if it happens in the future you could do an interactive rebase and tidy it up. This event system is super nice!

Aquarius/Src/Aquarius/Events/EventHandler.h
liamjwarren: Do you mean to use std::bind here or when we invoke it? I think you mean when we invoke it based on reading this [Article](https://en.cppreference.com/w/cpp/utility/functional/bind )

Burke-Daniel: When we invoke it yeah, this was just me thinking out loud haha

liamjwarren:
Thanks for the review guys! Just pushed a commit with the changes we discussed. I added in the following parameters for events: code which is used for key or mouse codes and xAxis & yAxis which is used for mouse scrolling and position. I also added in some getters for these in the events. I had to leave the static in front of the callback function definitions as when I removed it I was receiving a linking error. Let me know what you guys think and if you have any questions!

liamjwarren:
Once I get approvals I am planning on merging this branch and then making a new one to try and implement it as we discussed FYI.

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -64,11 +63,11 @@ namespace Aquarius {
...
-        glfwSetKeyCallback(m_Window, KeyboardEvent::Callback);
-        glfwSetCursorPosCallback(m_Window, MouseEvent::cursorPositionCallback);
-        glfwSetCursorEnterCallback(m_Window, MouseEvent::cursorInWindowCallback);
-        glfwSetMouseButtonCallback(m_Window, MouseEvent::mouseButtonCallback);
-        glfwSetScrollCallback(m_Window, MouseEvent::mouseScrollCallback);
+        glfwSetKeyCallback(m_Window, keyboardCallback);
+        glfwSetCursorPosCallback(m_Window, mouseCursorPositionCallback);
+        glfwSetCursorEnterCallback(m_Window, mouseCursorInWindowCallback);
+        glfwSetMouseButtonCallback(m_Window, mouseButtonCallback);
+        glfwSetScrollCallback(m_Window, mouseScrollCallback);
```
colton-smith: Very nice! Might still want to consider placing those guys in a callbacks:: namespace or similar though!

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,73 @@
...
+                    AQ_CORE_TRACE("key %v has been released.", key);
+                    break;
+
+                case GLFW_REPEAT:
+                    AQ_CORE_TRACE("key %v is being pressed repeatedly.", key);
+                    break;
+
+                default:
+                    break;
+            }
```
colton-smith: The callbacks have to emit events

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,54 @@
...
+    protected:
+        T m_Type;
+        bool m_Handled = false;
+
+        // The below are parameters that are used to provide more information about the event.
+        // If the parameter is not applicable to the event it will be set to 0.
+        // Code is used for Mouse codes & Key codes.
+        int m_Code;
+        double m_xAxis;
+        double m_yAxis;
```
colton-smith: These members belong in the derived event classes, not here

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,33 @@
...
+    {
+    public:
+        KeyPressedEvent() : Event<KeyboardEvents>(KeyboardEvents::KeyPressedEvent, m_Code, 0,0) {};
+    };
+
+    class KeyRepeatedEvent : public Event<KeyboardEvents>
+    {
+    public:
+        KeyRepeatedEvent() : Event<KeyboardEvents>(KeyboardEvents::KeyRepeatedEvent, m_Code, 0, 0) {};
+    };
```
colton-smith: Give these classes the private data members and getters/setters

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+#include "Aquarius/Core/Log.h"
+
+#include <GLFW/glfw3.h>
+
+
+namespace Aquarius {
+
```
colton-smith, line 13: Probably a good idea: Callbacks:: namespace

colton-smith (changes requested):
Will need to move data members out of the base event class and into the event sub classes, and add some getter/setter methods to the sub classes as well.

Also, consider adding a namespace to the callbacks and also adding event creation and emitting to the appropriate callbacks.

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "Event.h"
+
+#include <GLFW/glfw3.h>
+#include <unordered_map>
+#include <vector>
```
colton-smith: Cant get the engine to build without including <functional>

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,33 @@
...
+
+    enum class KeyboardEvents
+    {
+        KeyPressedEvent, KeyRepeatedEvent, KeyReleasedEvent,
+    };
+
+    class KeyPressedEvent : public Event<KeyboardEvents>
+    {
+    public:
+        KeyPressedEvent() : Event<KeyboardEvents>(KeyboardEvents::KeyPressedEvent, m_Code, 0,0) {};
```
colton-smith: This is kinda crazy, you are initializing a class with its own member? It seems cyclical in nature and very not sure how it works>

Aquarius/Src/Aquarius/Events/Event.h
Burke-Daniel: ^^ This will prevent the members from never being necessary, as they will only be present in events which contain them as information

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include <GLFW/glfw3.h>
```
Burke-Daniel: I don't think this needs to be included here

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "Event.h"
+
+#include <GLFW/glfw3.h>
```
Burke-Daniel: Ditto here

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
Burke-Daniel: Yeah I feel like this is very dangerous, Better to add the member variables in to these derived classes so that the callbacks can pass them in to these constructors when they create the appropriate event objects

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,42 @@
...
+            if((subscribers.find(event.getEventType())) == subscribers.end())
+                return;
+
+            for(auto&& subscriber : subscribers.at(event.getEventType()))
+            {
+                // Notify of event only if the event has not already been handled
+                if(!event.isHandled())
+                {
+                    subscriber(event);
+                }
```
Burke-Daniel, line 34: So to me, this looks like a client can set the handled member variable to true, and then nothing else occurs. If the event is handled, does it make sense to just break out of this for loop? This will just loop through the rest of the vector and do nothing if the event has already been handled by one of the other subscribers

Burke-Daniel (changes requested):
Couple more small things then I'd say you can merge after another look and go ahead and add the event creation to the callbacks, add event handlers to the application class, etc.

liamjwarren:
Alright, this is what I have gotten done to date. I still need to troubleshoot the event handler portion, but I now have the window and key callback’s now creating an event and logging this correctly. 

I will troubleshoot the event handler tomorrow; I need to find the best way to tie it into the application and make sure it actually works as I tried a few different ways tonight and was getting seg faults. Once I have this done, I think everything else should fall into place in terms of subscriptions and notifications. I am thinking it may be good for testing purposes to create some unit tests or assertions for the subscriptions and notification functions are working as intended.

Let me know if you see any glaring issues or if you have any questions. I will try and get another updated commit up tomorrow as well with tying in the event handler.

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,81 @@
...
+                }
+
+                default:
+                    break;
+            }
+        }
+
+        static void mouseScrollCallback(GLFWwindow *window, double xOffSet, double yOffset)
+        {
+            MouseScrolledEvent event(xOffSet, yOffset);
```
colton-smith: OffSet and offset

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,31 @@
...
+    enum class eventType
+    {
+        None = 0,
+        KeyPressedEvent, KeyRepeatedEvent, KeyReleasedEvent,
+        MouseButtonPressedEvent, MouseButtonReleasedEvent, MouseMovedEvent, MouseScrolledEvent, MouseInWindowEvent
+    };
+
+#define EVENT_TYPE(type) static eventType getStaticType() { return eventType::type; }\
+                                                            virtual eventType getEventType() const override { return getStaticType(); }\
+                                                            virtual const char* getName() const override { return #type; }
```
colton-smith: As much as this is useful, it may be okay to just write these all out. Also, we can probably do std::string and not const char pointer

Aquarius/Src/Aquarius/Events/Event.h
```diff
@@ -0,0 +1,31 @@
...
+    public:
+        virtual eventType getEventType() const = 0;
+        virtual const char* getName() const = 0;
+        bool m_Handled = false;
+        virtual ~Event() = default;
+
+        bool isHandled()
+        {
+            return m_Handled;
+        };
```
colton-smith: m_Handled should be private otherwise the getter doesn't really make sense

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,55 @@
...
+namespace Aquarius {
+
+    class EventHandler
+    {
+    public:
+        using functionType = std::function<void(const Event&)>;
+
+        EventHandler(Event& event) : m_event(event)
+        {
+            logCreation();
```
colton-smith: Not too sure we need to keep log messages in constructors in release / master

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        void logCreation()
+        {
```
colton-smith: I dont really mind this method tho, its essentially like a to_string method

Aquarius/Src/Aquarius/Events/EventHandler.h
colton-smith: @Burke-Daniel I think we should break when it gets handled for sure. However, some events may need information from an event but not actually consume the event, therefore we shouldnt break after line 33. Maybe we add a check after line 34, if (handled) {break;}

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,55 @@
...
+                {
+                    subscriber(event);
+                }
+            }
+        }
+
+        virtual ~EventHandler() = default;
+
+    private:
+        std::unordered_map<eventType, std::vector<functionType>> subscribers;
```
colton-smith: should be prefixed m_

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,67 @@
...
+#include <GLFW/glfw3.h>
+
+
+namespace Aquarius {
+
+    class KeyEvent : public Event
+    {
+    public:
+
+        int getCode()
```
colton-smith: This method should be const, and we should be using Input::KeyCode enum for the keycode member

Generally, any method that won't modify the class should be marked  const

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    class KeyPressedEvent : public KeyEvent
+    {
+    public:
+        KeyPressedEvent(int keyCode) : KeyEvent(keyCode) {}
+
+        EVENT_TYPE(KeyPressedEvent)
+
+        void logEvent()
```
colton-smith: Looks as tho every subclass of event has a logEvent class, maybe it should be declared virtual in the base event class?

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,67 @@
...
+        EVENT_TYPE(KeyRepeatedEvent)
+
+        void logEvent()
+        {
+            AQ_CORE_TRACE("KeyRepeatedEvent: %v", m_keyCode);
+        }
+
+    protected:
+        int m_repeatCount;
+    };
```
colton-smith, line 57: There is no way for an external caller to access this member because it is protected. I doubt this will be subclassed so repeat count could be private (with a getter to provide access).

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
...
+#include <GLFW/glfw3.h>
+
+
+namespace Aquarius {
+
+    class MouseMovedEvent : public Event
+    {
+    public:
+        MouseMovedEvent(float xCoordinate, float yCoordinate)
+                : m_XCoordinate(xCoordinate), m_YCoordinate(yCoordinate) {}
```
colton-smith: I think just m_x and m_y would suffice? Or x_pos? x_coord? xCoordinate seems a little bit verbose.

For me to get the x/y of a mouse, I would need to call:
```
mouseMovedEvent.getXCoordinate()
mouseMovedEvent.getYCoordinate()
```

colton-smith: I would also add a getter that returns a glm::vec2 or std::pair of the (x,y) coordinates

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
...
+
+        float getXCoordinate()
+        {
+            return m_XCoordinate;
+        }
+
+        float getYCoordinate()
+        {
+            return m_YCoordinate;
+        }
```
colton-smith: const

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
...
+
+    private:
+        float m_XCoordinate;
+        float m_YCoordinate;
+    };
```
colton-smith: Maybe make the member names better reflect a mouse scroll, this is a copy+paste of the MouseMovedEvent

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
...
+    class MouseButtonEvent : public Event
+    {
+    public:
+        int getCode()
+        {
+            return m_buttonCode;
+        }
+
+    protected:
+        int m_buttonCode;
```
colton-smith: Use the enum in Input.h instead of ints again

Aquarius/Src/Aquarius/Events/WindowEvent.h
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Aquarius {
+/*
+    enum class WindowEvents
+    {
+        WindowClosedEvent, WindowResizedEvent, WindowFocusedEvent, WindowMovedEvent
+    };
+
+    // Events will be added here as they are needed.
+*/
```
colton-smith: window resized will be needed ASAP

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+
+        int inWindow()
+        {
+            return m_inWindow;
+        }
+
+    private:
+        int m_inWindow;
```
colton-smith: Should probably be a bool

Aquarius/Src/Aquarius/Events/WindowEvent.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "Event.h"
+
+#include <GLFW/glfw3.h>
+
```
colton-smith: Probably wont need this include

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,135 @@
+#pragma once
+
+#include "Event.h"
+
+#include <GLFW/glfw3.h>
+
```
colton-smith: Dont think this is used

colton-smith (changes requested):
Lots of small stuff to be changed, lmk if you have any questions

Aquarius/Src/Aquarius/Core/Application.cpp
```diff
@@ -33,6 +34,11 @@ namespace Aquarius {
 		}
 	}
 
+	void Application::onEvent(Event &event)
+	{
+	    EventHandler handler(event);
```
Burke-Daniel: So it seems like any time onEvent is called from some piece of code (glfw callback functions for example), a new EventHandler instance is created, here (i.e constructor is called) and then it will immediately go out of scope. Do we potentially want to make EventHandler a member of the Application class, and that way then pieces of code that want to subscribe to certain events can access the handler through the static Application instance. Or I guess the alternative here is that we let the client handler the event stuff in their derived application class (though I assume we are going to want to have an EventHandler to reference through the glfw callback functions

Aquarius/Src/Aquarius/Events/Callbacks.h
```diff
@@ -0,0 +1,81 @@
...
+                    event.logEvent();
+                    break;
+                }
+
+                case GLFW_REPEAT:
+                {
+                    KeyRepeatedEvent event(key,1);
+                    event.logEvent();
+                    break;
+                }
```
Burke-Daniel: So here for example, are we gonna want to have something like:
```C++
KeyPressedEvent event(key);
event.logEvent();
Application::get()->onEvent(event);
break;
```
where the call to Application::onEvent would then call it's event handler's `notify` method?

Aquarius/Src/Aquarius/Events/EventHandler.h
Burke-Daniel: Can probably wrap in `#ifdef AQ_DEBUG` or something

Aquarius/Src/Aquarius/Events/EventHandler.h
```diff
@@ -0,0 +1,55 @@
...
+        Event& m_event;
```
Burke-Daniel: Does EventHandler need to have reference to a particular Event? Cause notify is having an event passed into it. Only place this is used is on construction, though I suppose maybe derived classes will use it or something. To avoid having to pass in an Event to construct the Handler, what if m_event was instead an Event*?

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,67 @@
...
+
+        int getCode()
+        {
+            return m_keyCode;
+        }
+
+    protected:
+        int m_keyCode;
+
+        KeyEvent(int keyCode) : m_keyCode(keyCode) {}
```
Burke-Daniel: So if you use the Input::KeyCodes instead of ints here, you can just `static_cast<Input::KeyCode>(keyCode)` here and that should work

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
Burke-Daniel: ^^ I agree, probably don't make it purely virtual though so that the derived class doesn't have to override it (could just be a virtual function with an empty body)

Burke-Daniel (changes requested):
Ok I can see this coming together better now, couple questions and couple comments for clarity, but main thing I am wondering is if you have decided to follow the observer pattern more tightly where the EventHandler object distributes the events to different places through the functions in the map, or more localized where there is a chain of event handling

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
colton-smith: To convert glfw to AQ keycodes?

Aquarius/Src/Aquarius/Events/EventHandler.h
colton-smith: Why does the handler need an event at all? With this architecture, we shouldn't need that idt

Aquarius/Src/Aquarius/Events/Callbacks.h
colton-smith: Thats what I picture

Aquarius/Src/Aquarius/Core/Application.cpp
colton-smith: I didnt even see this change, I thought the Application class would just have one event handler instance

Aquarius/Src/Aquarius/Events/EventHandler.h
Burke-Daniel: Yeah that's why I was a bit confused about which direction we were going with the architecture of this

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
Burke-Daniel: Yeah

Aquarius/Src/Aquarius/Events/MouseEvent.h
liamjwarren: It has to be an int due to the GLFW callback, Daniel mentioned this before and thought it was something to do with C. That being said I have updated the inWindow() method to return a bool value rather than an int which could be used to check if the cursor is in the window. I will have this updated in my commit this evening.

colton-smith: I mean you could cast the int returned by the callback to a bool before creating this event right? Doesn't really matter though

```
        static void cursorInWindowCallback(GLFWwindow* window, int inWindow)
        {
            bool bInWindow = inWindow;
            if (bInWindow)
            {
                AQ_INFO("Cursor has entered application window.");
                MouseEvent e(bInWindow)
            }

            else
            {
                AQ_INFO("Cursor has left application window.");
            }
        }
```

Its implicitly converted as well

liamjwarren: Nvm, did a cast.

liamjwarren:
Alright, here is my latest commit for this PR. A few things to note that are now implemented:
- logEvent() is now a virtual method
- Codes for Keyboard & Mouse events are now from the enums in Input.h
- Window Resize event and callback
- EventHandler updated based on comments
- Callbacks are in a Callbacks namespace
- Application has one instance of EventHandler which onEvent(Event &event) will not invoke the notify method of the event handler
- Callbacks are tied in to trigger onEvent now  

Let me know what you guys think!

Aquarius/Src/Aquarius/Core/Window.h
```diff
@@ -2,6 +2,7 @@
 
 #include "Aquarius/Core/Utility.h"
 #include "Aquarius/Renderer/RenderingContext.h"
+#include "Aquarius/Events/Event.h"
```
colton-smith: I dont see where this include is used

Aquarius/Src/Aquarius/Events/MouseEvent.h
```diff
@@ -0,0 +1,133 @@
...
+                : m_XPosition(xPosition), m_YPosition(yPosition) {}
+
+        EVENT_TYPE(MouseMovedEvent)
+
+        void logEvent() override
+        {
+            AQ_CORE_TRACE("MouseMovedEvent: %v, %v", m_XPosition, m_YPosition);
+        }
+
+        float getXCoordinate()
```
colton-smith: Maybe match the field name and getter name?

colton-smith (changes requested):
Super close to GTM:
- Need to update the getters and log methods to be const
- Need to break when the event is handled to avoid unnecessarily iteration
- Subscribers field in event handler should be prefixed m_ 

Then you should be g2g

Aquarius/Src/Aquarius/Core/Window.h
liamjwarren: Oh Liam...

Aquarius/Aquarius.h
```diff
@@ -16,9 +16,12 @@
 #include "Aquarius/Renderer/Shader.h"
 
 // -- Events
-#include "Aquarius/Events/Keyboard.h"
-#include "Aquarius/Events/Mouse.h"
-#include "Aquarius/Events/Window.h"
+#include "Aquarius/Events/Callbacks.h"
```
colton-smith: This may be an exception - I doubt this needs to be exposed to the public

colton-smith (approved):
I added one note, not sure we need to have Callbacks.h in Aquarius.h as no one will need to use those except the Window class (everyone else can then sub to events).

GTM after @Burke-Daniel has a comb through, nice work!

Aquarius/Src/Aquarius/Events/KeyboardEvent.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include "Event.h"
+
```
Burke-Daniel, line 6: So I just wanted to test functionality of this, and it works but I had to add:
```C++
#include "Aquarius/Core/Input.h"
#include "Aquarius/Core/Log.h"
```
here.  But what I did to test was change using the escape key to close the window to be the application subscribing to a keyPressedEvent instead of using input polling. This is how I did that:
```C++
Application::Application(std::string&& windowName)
		: m_Window(Window::Create(800, 600, std::move(windowName)))
	{
		s_Application = this;

		Log::initLoggers();
		m_Window->Initialize();

		m_EventHandler.subscribe(eventType::KeyPressedEvent,
			[&](const Event& event)
			{
				try
				{
					const auto& keyEvent = dynamic_cast<const Aquarius::KeyPressedEvent&>(event);
					if (keyEvent.getCode() == Aquarius::Input::KeyCode::Key_escape)
					{
						glfwSetWindowShouldClose(m_Window->get(), true);
					}
				}
				catch (std::bad_cast&)
				{
					// Wrong event type, don't need to handle it
				}
			});

		AQ_CORE_INFO("Window Initialized Successfully");
	}
```
wIth a cheeky little lambda, and dynamic cast to cast the base Event reference down to a KeyPressedEvent so that I could call `getCode()`, this works to close the window as well, which is pretty sick!

Burke-Daniel (approved):
This is great! See my one comment about a couple of includes in KeyboardEvent.h and then GTM

PR #34, opened 2021-03-06
Quad Rendering

colton-smith:
Quad rendering... Sort of.

This PR adds some static structs to the renderer namespace that contain information about the scene:
- View matrix (Camera view)
- Projection matrix  (Ortho / Perspective)
- Vertex Array configured for rendering quads on Init
- Generic shader with model, view, and projection uniforms 

A few methods were added to set the view and matrix projections (view will be done via a camera most likely), as well as draw a quad.

Quads just reuse the same Vertex array defined statically, with a different model matrix calculated based on the desired position rotation, and size. Size is a pixel value.  

It works, rendering the quad at the correct rotation with the top left corner at the (x,y) position specified. Once we get textures I can confirm a theory I have that it is rendering upside down (nothing some texture coords cant fix). This is because the screen space coords start at 0,0 at the top left, while the NDC vertices start at 0,0 at the origin. It was a little hard to wrap my head around getting the quad to convert properly from the vertex positions to pixels on the screen but I think its at a good enough place to get in there. 

image


liamjwarren (approved):

Aquarius/Src/Aquarius/Renderer/Renderer.h
```diff
@@ -17,7 +17,14 @@ namespace Aquarius {
...
         void ClearColor(glm::vec3 color);
 
+        void SetView(const glm::mat4& view);
+        void SetProjection(const glm::mat4& projection);
+
         void Submit(VertexArray* vertexArray, Shader* shader);
+        void DrawQuad(const glm::vec2& pos, 
+                      const glm::vec2& size, 
+                      float rotationDegrees = 0.0f, 
+                      const glm::vec4& color = { 0.0f, 0.0f, 0.0f, 1.0f });
```
Burke-Daniel, line 27: Lovely use of default args here

Aquarius/Src/Aquarius/Renderer/Shader.cpp
```diff
@@ -131,13 +131,13 @@ namespace Aquarius {
...
         int location = glGetUniformLocation(m_ID, name.c_str());
-        glUniform4i(location, values.x, values.y, values.z, values.w);
+        glUniform4f(location, values.x, values.y, values.z, values.w);
     }
 
     void Shader::setMat4(const std::string& name, bool transpose, const glm::mat4& value) const
     {
         int location = glGetUniformLocation(m_ID, name.c_str());
-        glUniformMatrix4fv(location, 1, transpose, glm::value_ptr(value));
+        glUniformMatrix4fv(location, 1, transpose ? GL_TRUE : GL_FALSE, glm::value_ptr(value));
```
Burke-Daniel, line 140: Ahhh will ya look at that... nice lil ternary lmao

Burke-Daniel (approved):
This is sick! I need to catch up on openGL but I understand mostly all of this, GTM

PR #35, opened 2021-03-06
Generate Visual Studio Project Files w/ CMake

colton-smith:
Quick PR making the lives of Visual Studio users a little easier - generate filters that match the actual src tree with cmake. I find it nicer to have it organized like this instead of "Header/" and "Source/" like the default filters. 

LMK what you guys think, and @liamjwarren definitely test it on mac / CLion. Worried it might break on MAC seeing how visual studio isn't a thing there. I can probably add an "IF MSVC" or something


liamjwarren (approved):
Ran it on CLion on Mac and no issues, blue window opens with no build errors.

Burke-Daniel (approved):
This is very nice! Seems like it works for Liam on Mac/CLion so GTM

PR #36, opened 2021-03-06
Texture Abstraction Class

Burke-Daniel:
This PR adds a Texture class, which uses stbi to abstract away the process of creating textures. Up to a minimum of 16 textures can be created, with the ability to set the texture wrapping and texture filtering options, and to bind and unbind individual textures. The sandbox example has also been modified to render the container image from the learnOpenGL tutorial to the window.


Aquarius/Src/Aquarius/Renderer/Texture.cpp
```diff
@@ -0,0 +1,59 @@
...
+		generateTexture(texture, hasAlpha);
+	}
+
+	bool Texture::generateTexture(const std::string& texture, bool hasAlpha)
+	{
+		bool success = false;
+
+		glGenTextures(1, &m_ID);
+		glBindTexture(GL_TEXTURE_2D, m_ID);
+		m_TextureUnitNum = GL_TEXTURE0 + s_NextTextureUnitNum++;
```
colton-smith: Nice

Aquarius/Src/Aquarius/Renderer/Texture.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, static_cast<int>(m_Configuration.textureWrapS));
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, static_cast<int>(m_Configuration.textureWrapT));
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, static_cast<int>(m_Configuration.minTextureFiltering));
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, static_cast<int>(m_Configuration.magTextureFiltering));
+		unsigned char* data = stbi_load(texture.c_str(), &m_Width, &m_Height, &m_NumColorChannels, 0);
```
colton-smith, line 29: Prior to calling this, the following should be called somewhere, once:
```
 stbi_set_flip_vertically_on_load(true);
 ```

File: Sandbox/Shaders/fragmentShader.glsl [hunk 3]
```diff
@@ -3,8 +3,11 @@
...
 in vec3 fColor;
+in vec2 fTexCoord;
+
+uniform sampler2D fTexture;
 
 void main()
 {
-     FragColor = vec4(fColor, 1.0);
+     FragColor = texture(fTexture, fTexCoord) * vec4(fColor, 1.0); 
 }
```
colton-smith: Is it cool if we make another set of vertex/frag shaders? I think we should have one that is pliable and then one thats static and used in the quadRenderer

colton-smith: For the quad renderer, I'm thinking the texture coords will just be uniforms and not vertex data

Aquarius/Src/Aquarius/Renderer/Texture.cpp
```diff
@@ -0,0 +1,59 @@
...
+		
+		if (data != nullptr)
+		{
+			int format = hasAlpha ? GL_RGBA : GL_RGB;
+
```
colton-smith: Might be better to use uint32_t here for consistency

colton-smith: To avoid narrowing conversions, GLenum is 32 bits

colton-smith (changes requested):
Apart from a couple of small things I noted in the comments:

What happens if a texture object is destroyed? I think that textures unit number will be "lost" so to speak. 

 Might be nice to somehow separate texture asset creation from managing the texture slots

Sandbox/Shaders/fragmentShader.glsl
colton-smith: Also, I was wondering if the quad renderer shader should be moved in-engine

Burke-Daniel: Added default and modifiable shaders to the Shaders folder in Sandbox, I'll make a comment on your PR to move your changes to the shader for quad rendering to within the engine

Aquarius/Src/Aquarius/Renderer/Texture.cpp
Burke-Daniel: Good call

Burke-Daniel:
Just added a couple of commits addressing your comments, I'll get you to take another look @colton-smith

Aquarius/Src/Aquarius/Renderer/Texture.cpp
```diff
@@ -0,0 +1,57 @@
...
+		stbi_image_free(data);
+		glBindTexture(GL_TEXTURE_2D, 0);
+
+		return success;
+	}
+
+	void Texture::bind(uint32_t textureSlot) const
+	{
+		glActiveTexture(textureSlot);
```
colton-smith: Maybe this should be like, GL_TEXTURE0 + textureSlot?

Aquarius/Src/Aquarius/Renderer/Texture.h
```diff
@@ -0,0 +1,59 @@
...
+		int getHeight() const { return m_Height; }
+		int getNumColorChannels() const { return m_NumColorChannels; }
+
+		void setConfiguration(const TextureConfiguration& configuration) { m_Configuration = configuration; }
+
+	private:
+		uint32_t m_ID = 0;
+		int m_Width = 0;
+		int m_Height = 0;
+		int m_NumColorChannels = 0;
```
colton-smith, line 56: I dont think this is ever set ?

Sandbox/Sandbox.cpp
```diff
@@ -16,24 +16,31 @@ void SandboxLayer::onCreation()
...
+	m_texture = std::make_shared<Aquarius::Texture>("Assets/container.jpg");
+	m_texture->bind(GL_TEXTURE0);
 }
 
 void SandboxLayer::onUpdate(Aquarius::timeDelta_t)
 {
 	Aquarius::Renderer::ClearColor({ 0.2, 0.3, 0.7 });
 	Aquarius::Renderer::Clear();
+
+	m_texture->bind(GL_TEXTURE0);
```
colton-smith: I think it makes sense to pass this an offset from GL_TEXTURE0

colton-smith (changes requested):
Couple of small things then we are GTM this!

- color channel member is never set
- We should just provide an int offset to bind, then bind will just use that as an offset from GL_TEXTURE0 -> this way calling code doesn't need the GL_TEXTURE enum.

Aquarius/Src/Aquarius/Renderer/Texture.h
Burke-Daniel: It gets set by the call to `stbi_load` in `generateTexture`

colton-smith: Jesus i commented on that previously as well

colton-smith (approved):
GTM

Aquarius/Src/Aquarius/Renderer/Texture.cpp
```diff
@@ -0,0 +1,61 @@
...
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, static_cast<int>(m_Configuration.textureWrapT));
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, static_cast<int>(m_Configuration.minTextureFiltering));
+		glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, static_cast<int>(m_Configuration.magTextureFiltering));
+		unsigned char* data = stbi_load(texture.c_str(), &m_Width, &m_Height, &m_NumColorChannels, 0);
+		
+		if (data != nullptr)
+		{
+			uint32_t format = hasAlpha ? GL_RGBA : GL_RGB;
+
+			glTexImage2D(GL_TEXTURE_2D, 0, format, m_Width, m_Height, 0, format, GL_UNSIGNED_BYTE, data);
```
colton-smith: Technically the first format argument is the internal format and should be RGBA8 or RGB8 (RGBA works only for backwards compatibility according to youtube)

colton-smith (changes requested):
- Destructor should call glDeleteTextures
- 2nd argument of glTexImage2D is the internal format and should have a number attached which is the bits per pixel

Aquarius/Src/Aquarius/Renderer/Texture.cpp
Burke-Daniel: Ahh I knew they were different but in learnopenGL they used GL_RGB and RGBA for both, but this is an easy fix

colton-smith (approved):
send it

PR #37, opened 2021-03-09
Add Orthographic Camera

colton-smith:
## Summary
This PR adds a general purpose orthographic camera that is used to set the view and projection matrices of the scene. The camera can pan and zoom with the use of the WASD and UP/DOWN keys repsectively. 

In the future I will likely refactor this to inherit from a camera class, allowing me to use Camera object in the quad renderer instead of OrthographicCamera - helping me to satisfy the DI and OC principles of software design 😂

A lot of the changes that are in here are from the quad rendering PR, as I had to rebase on there to get that stuff. 

When we merge that first, will the diff of this PR shrink? @Burke-Daniel @liamjwarren 
 
## Key Files
 - `OrthographicCamera.cpp`
 - `OrthographicCamera.h`
 - `Renderer.cpp`
 - `Sandbox.cpp`


Aquarius/Aquarius.h
```diff
@@ -16,6 +16,7 @@
 #include "Aquarius/Renderer/VertexArray.h"
 #include "Aquarius/Renderer/Shader.h"
 #include "Aquarius/Renderer/Texture.h"
+#include "Aquarius/Renderer/OrthographicCamera.h"
```
liamjwarren: This should be at the top of the includes to be in alphabetical order.

Aquarius/Aquarius.h
```diff
@@ -16,6 +16,7 @@
 #include "Aquarius/Renderer/VertexArray.h"
```
liamjwarren: This should be at the bottom of the includes to be in alphabetical order.

Aquarius/Src/Aquarius/Renderer/OrthographicCamera.cpp
```diff
@@ -0,0 +1,85 @@
...
+        // Zoom 
+        if (Input::isKeyPressed(Input::KeyCode::Key_down))
+        {
+            m_Zoom = m_Zoom - dzoom;
+        }
+
+        if (Input::isKeyPressed(Input::KeyCode::Key_up))
+        {
+            m_Zoom += dzoom;
+        }
```
Burke-Daniel, line 58: Is it better to use Events for this now or do people typically use input polling for this? I spose it works either way so It doesn't matter

Burke-Daniel: Also when I was moving the quad around I noticed that the controls were inverted, was that intentional or

Aquarius/Src/Aquarius/Renderer/Renderer.cpp
```diff
@@ -66,6 +66,13 @@ namespace Aquarius {
                                                              "Aquarius/Src/Aquarius/Renderer/Shaders/fragmentShader.glsl");
             }
 
+            void BeginScene(OrthographicCamera* camera)
```
Burke-Daniel, line 69: Could you begin a scene without a camera (i.e. passing in a nullptr here)? This could be a reference potentially

Sandbox/Sandbox.cpp
```diff
@@ -3,52 +3,78 @@
...
-		Aquarius::VertexElement(Aquarius::ShaderType::Float, 2, false)
-	});
-
-	m_vertexArray = std::make_shared<Aquarius::VertexArray>(m_vertexBuffer, m_indexBuffer, *m_bufferLayout);
-	m_vertexArray->activate();
-	Aquarius::Renderer::Init();
-	m_texture = std::make_shared<Aquarius::Texture>("Assets/container.jpg");
-	m_texture->bind(0);
+    m_ShaderProgram = std::make_shared<Aquarius::Shader>("Sandbox/Shaders/modifiableVertexShader.glsl", 
+                                                         "Sandbox/Shaders/modifiableFragmentShader.glsl");
```
Burke-Daniel, line 12: Ok nice so these shaders are for playing around with and now the quad renderer shader is in the engine. Sweeet

Sandbox/Sandbox.cpp
```diff
@@ -3,52 +3,78 @@
...
+        Aquarius::VertexElement(Aquarius::ShaderType::Float, 2, false),
+        Aquarius::VertexElement(Aquarius::ShaderType::Float, 3, false),
+        Aquarius::VertexElement(Aquarius::ShaderType::Float, 2, false)
+    });
+
+    m_vertexArray = std::make_shared<Aquarius::VertexArray>(m_vertexBuffer, m_indexBuffer, *m_bufferLayout);
+    m_vertexArray->activate();
+
+    Aquarius::Application* app = Aquarius::Application::get();
+    Aquarius::Window* window = app->getWindow();
```
Burke-Daniel, line 29: This is satisfying to me for some reason haha, I find it cool how we have global access to the application class. Idk why everybody be hating on globals/singletons

Sandbox/Sandbox.cpp
```diff
@@ -3,52 +3,78 @@
...
+    int height = window->getHeight();
+    int width = window->getWidth();
+
+    m_Camera = std::make_shared<Aquarius::OrthographicCamera>(1, 0.01, height, width);
+
+    Aquarius::Renderer::Init();
+
+    m_texture = std::make_shared<Aquarius::Texture>("Sandbox/Assets/container.jpg");
+    m_texture->bind(0);
+    Aquarius::Renderer::Init();
```
Burke-Daniel, line 39: This doesn't need to be here twice I'm assuming

Sandbox/Sandbox.cpp
```diff
@@ -3,52 +3,78 @@
...
 }
 
-void SandboxLayer::onUpdate(Aquarius::timeDelta_t)
+void SandboxLayer::onUpdate(Aquarius::timeDelta_t ts)
```
Burke-Daniel, line 42: Glad to see the timeDelta came in handy now

Sandbox/Sandbox.cpp
```diff
@@ -3,52 +3,78 @@
...
+    );
+
+    Aquarius::Renderer::DrawQuad(
+        {100, 150},
+        {100, 100},
+         0,
+        {1, 1, 1, 1}
+    );
 
+    /// Aquarius::Renderer::Submit(m_vertexArray.get(), m_ShaderProgram.get());
```
Burke-Daniel, line 67: Oooook this makes more sense to me now, you don't need to call Submit after DrawQuad because DrawQuad does the same operations as Submit, makes sense.

Burke-Daniel (approved):
Nothing really major here that I'm wondering about except for the keyboard controls controlling the quad backwards, this works for me so GTM after I'd say.

P.S. I am starting to understand the renderer much better now haha

liamjwarren (approved):
Very cool! Just two minor comments on include order otherwise GTM!

Aquarius/Src/Aquarius/Renderer/OrthographicCamera.cpp
colton-smith: Input polling is fine I'd say, and we discussed the camera movement

Sandbox/Sandbox.cpp
colton-smith: Yeah UE4 has a similar thing with a class called  GEngine

Sandbox/Sandbox.cpp
colton-smith: highly useful

Sandbox/Sandbox.cpp
colton-smith: Yeah submit is like DrawQuad except you supply your own vertex data and shader

Sandbox/Sandbox.cpp
colton-smith: Yeah we can do this temporarily, not sure how to handle shaders tbh hahaha

PR #47, opened 2021-03-22
Add Boolean Active Flag to Layer

colton-smith:
## TINY PR ALERT

This PR just adds a boolean flag to the base layer class, and modifies the engine loop to check if a layer is active before updating it. Any layer handling events will have to add something similar to the top of their callback they are registering:

```
if (not isActive())
{
    return;
}
```


Burke-Daniel (approved):
GTM

PR #48, opened 2021-03-22
PONG!!!

Burke-Daniel:
Alright ladies and gentleman, the most sophisticated pong game you have ever seen... ok well I think it is half decent. But anyways, if we wanna discuss then maybe we can have a quick call, some main things to note:

Three methods of paddle control:
1. Keyboard
2. Mouse
3. AI

- Ball and Paddle are structs because I didn't wanna write getters and setters (I will change if we want to properly encapsulate)
- Escape pauses the game

There are still quite a few things we can add to refine this for final submission, if you guys have any ideas lmk


Burke-Daniel:
Just fixed a bug with collision detection lads, before the collision detection was only using the top left/right y coordinate of the ball, so the ball could go through the paddle if it only partially collided, but now it uses both the top and bottom left/right y coordinate. Also I implemented pause separate from whether or not the layer is active, so that the pausing is persistent (before if you paused then switched layers using the manager layer, when you switched back it would unpause the game)

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -101,10 +101,6 @@ namespace Aquarius {
 
     void Window::OnUpdate()
     {
-        if (Input::isKeyPressed(Input::KeyCode::Key_escape))
```
liamjwarren, line 104: Ahh this should probably be put back in for the application to exit.

Sandbox/PongLayer.cpp
```diff
@@ -0,0 +1,235 @@
...
+		auto keyPressEvent = static_cast<const Aquarius::KeyPressedEvent&>(event);
+		switch (keyPressEvent.getCode())
+		{
+			case(Aquarius::Input::KeyCode::Key_p):
+			{
+				m_isPaused = !m_isPaused;
+				break;
+			}
+
+			case(Aquarius::Input::KeyCode::Key_h):
```
liamjwarren, line 100: One thing I am wondering is if hyper mode should be activated until the user presses H again. Right now it resets once someone has scored and goes back to normal mode. I do not know if this is possible but it might be something we can refactor for the final submission.

Aquarius/Aquarius.h
```diff
@@ -20,6 +20,7 @@
 #include "Aquarius/Renderer/VertexBuffer.h"
 #include "Aquarius/Renderer/VertexArray.h"
 
+
```
liamjwarren: Extra line here

liamjwarren (commented):
Looks good, just a few comments.

colton-smith (approved):
GTM

PR #49, opened 2021-03-23
Layer management syst.... I mean layer

Burke-Daniel:
Hot swappable layers. That's it.

You can either use the pg up/pg down keys, or f/b to move forwards and backwards through layers, it will also cycle so if you try to go "up" on the last layer in the list, it will bring you back to the first layer.

Note there are pong layer changes in here that I made, but then cherry picked so they should also be on the pong branch.

### DIFF
--- a/Sandbox/ManagerLayer.cpp
+++ b/Sandbox/ManagerLayer.cpp
@@ -41,6 +41,12 @@ void ManagerLayer::onEvent(const Aquarius::Event& event)
 				m_ManagedLayers[m_ActiveLayerIndex]->deactivate();
 				m_ManagedLayers[++m_ActiveLayerIndex]->activate();
 			}
+			else
+			{
+				m_ManagedLayers[m_ActiveLayerIndex]->deactivate();
+				m_ActiveLayerIndex = 0;
+				m_ManagedLayers[m_ActiveLayerIndex]->activate();
+			}
 			break;
 		}
 
@@ -51,6 +57,12 @@ void ManagerLayer::onEvent(const Aquarius::Event& event)
 			{
 				m_ManagedLayers[m_ActiveLayerIndex]->deactivate();
 				m_ManagedLayers[--m_ActiveLayerIndex]->activate();
+			}
+			else
+			{
+				m_ManagedLayers[m_ActiveLayerIndex]->deactivate();
+				m_ActiveLayerIndex = m_ManagedLayers.size() - 1;
+				m_ManagedLayers[m_ActiveLayerIndex]->activate();
 			}
 			break;
 		}
--- a/Sandbox/ManagerLayer.cpp
+++ b/Sandbox/ManagerLayer.cpp
@@ -29,42 +29,47 @@ void ManagerLayer::onCreation()
 
 void ManagerLayer::onEvent(const Aquarius::Event& event)
 {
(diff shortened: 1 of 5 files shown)


Sandbox/ManagerLayer.cpp
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if (!m_ManagedLayers.empty())
+	{
+		m_ManagedLayers.front()->activate();
+		m_ActiveLayerIndex = 0;
+	}
+}
+
+void ManagerLayer::onEvent(const Aquarius::Event& event)
```
liamjwarren, line 30: We need the esc to exit out of the application, with the mouse disabled it makes it difficult to navigate and close the window.

liamjwarren (changes requested):
Looks good, just the esc key for easy exiting of the application.

Sandbox/ManagerLayer.cpp
colton-smith: @Burke-Daniel Esc key callback in window should handle this right?

Sandbox/ManagerLayer.cpp
```diff
@@ -0,0 +1,70 @@
...
+				m_ManagedLayers[++m_ActiveLayerIndex]->activate();
+			}
+			else
+			{
+				m_ManagedLayers[m_ActiveLayerIndex]->deactivate();
+				m_ActiveLayerIndex = 0;
+				m_ManagedLayers[m_ActiveLayerIndex]->activate();
+			}
+			break;
+		}
```
colton-smith: indentation is a little off here

Sandbox/PongLayer.cpp
```diff
@@ -135,8 +141,8 @@ void PongLayer::onUpdate(Aquarius::timeDelta_t dt)
 	m_RightPaddle.Render();
 	m_Ball.Render();
 
-	AQ_INFO("Frametime: %v ms", dt);
-	AQ_INFO("FPS: %v", 1000 / dt);
+	AQ_TRACE("Frametime: %v ms", dt);
+	AQ_TRACE("FPS: %v", 1000 / dt);
```
colton-smith: Where do we want to set the log level for the demo application?

Sandbox/Sandbox.cpp
```diff
@@ -91,7 +93,14 @@ void SandboxLayer::onUpdate(Aquarius::timeDelta_t ts)
...
 {
-    PushLayer(std::make_unique<PongLayer>());
+    Aquarius::Log::setLogLevel(Aquarius::LogLevel::Info);
+
+    auto testLayer1 = PushLayer(std::make_unique<SandboxLayer>());
+    auto testLayer2 = PushLayer(std::make_unique<PongLayer>());
+
+    PushLayer(std::make_unique<ManagerLayer>(
+        std::vector<Aquarius::Layer*>{ testLayer1, testLayer2 }));
+
```
colton-smith, line 103: Nice!

colton-smith (approved):
Literally just indentation in the switch case of ManagerLayer::OnEvent is warped, then GTM

Sandbox/ManagerLayer.cpp
liamjwarren: Yep, after reviewing this PR I went and did the pong one and made a note to put the esc back in the window haha

liamjwarren (approved):

PR #50, opened 2021-03-23
Intro Layer

liamjwarren:
Nice little intro layer for our demo simulations. I have removed the use of sleeping the thread on each update loop to instead increment the alpha by a smaller value.


liamjwarren:
Also noteworthy in this PR is the addition of a colour field in the textured draw quad method.

Burke-Daniel (approved):
GTM

PR #54, opened 2021-04-02
Refine Bubble Sort Layer

liamjwarren:
- Tied the delay in between on update calls as we did for the intro layer so that the bubble sort will be rendered clear regardless of computer.
- Can now safely pause the bubble sort simulation without the risk of causing a seizure! 
- TODO: I am going to try and get it so that the "winning" bar is green or some colour.


Sandbox/BubbleSortLayer.cpp
```diff
@@ -28,31 +28,34 @@ void BubbleSortLayer::onEvent(const Aquarius::Event& event)
...
             break;
-        }
+            }
     }
 }
 
-void BubbleSortLayer::onUpdate(Aquarius::timeDelta_t ts)
+void BubbleSortLayer::onUpdate(Aquarius::timeDelta_t dt)
 {
+    delay = dt * 750;
```
Burke-Daniel: You could make `delay` equal to 750 in the class definition similar to what you did before, and then create a local variable that is equal to dt * delay and use that when you pause the thread. This might make it easier to change the delay at runtime if you end up doing something like that through imGUI

Sandbox/BubbleSortLayer.cpp
```diff
@@ -63,7 +66,13 @@ void BubbleSortLayer::onUpdate(Aquarius::timeDelta_t ts)
         resetSort = false;
     }
 
-    if (isActive())
+    if(pauseSort)
```
Burke-Daniel: space between if and (

Sandbox/BubbleSortLayer.cpp
```diff
@@ -75,10 +84,10 @@ void BubbleSortLayer::onUpdate(Aquarius::timeDelta_t ts)
         j += 1;
 
         if(i == numRectangles - 1 && j == numRectangles)
```
Burke-Daniel: Same thing as above, just for consistency

Sandbox/BubbleSortLayer.cpp
```diff
@@ -28,31 +28,34 @@ void BubbleSortLayer::onEvent(const Aquarius::Event& event)
...
 
-        case (Aquarius::Input::KeyCode::Key_r): {
-            resetSort = true;
+            else
+            {
+                pauseSort = false;
+            }
             break;
-        }
+            }
```
Burke-Daniel: Indentation of brackets looks a bit messed up here

Burke-Daniel (approved):
Couple small style things and one suggestion, then GTM

Sandbox/BubbleSortLayer.cpp
```diff
@@ -28,31 +28,34 @@ void BubbleSortLayer::onEvent(const Aquarius::Event& event)
...
-    switch (keyPressEvent.getCode()) {
-        case (Aquarius::Input::KeyCode::Key_up): {
-            if (delay >= 0) {
-                delay = delay - 200;
-            }
+    switch (keyPressEvent.getCode())
+    {
+        case (Aquarius::Input::KeyCode::Key_r):
+            {
+            resetSort = true;
```
colton-smith, line 44: I would tab this line in

Sandbox/BubbleSortLayer.cpp
colton-smith: I second this, it would be very nice to allow this to be changed via ImGUI

Sandbox/BubbleSortLayer.cpp
```diff
@@ -63,7 +66,13 @@ void BubbleSortLayer::onUpdate(Aquarius::timeDelta_t ts)
         resetSort = false;
     }
 
-    if (isActive())
+    if(pauseSort)
+    {
+        Aquarius::Renderer::Clear();
```
colton-smith: I dont know why you need multiple different clear calls in the same onUpdate method

colton-smith (changes requested):
Couple of small changes then GTM

liamjwarren:
Changes have been made, @colton-smith . Back to you.

colton-smith (approved):
GTM

PR #57, opened 2021-04-04
ImGUI Integration

colton-smith:
## Brief

This PR adds the ImGUI library (tag v1.82) to the project. I removed all files we don't need, leaving the useful examples, demos, and documentation. This is a pretty good solution, the only issue is it makes it hard to bump the version (not likely we will need to do this in the next 3 days ahahah)

Using the example they give for glfw/OpenGL3, imgui is integrated with the application and layering system. ImGUI is initialized just after the window in the Application constructor. Each engine loop, a new ImGUI frame is created, then onUpdateGUI is called on each layer, then imgui::render is called. I think callbacks are being hooked up behind the scenes because mouse and keyboard navigation of the demo works very well. The only thing that doesnt work (not sure it should work) is the file explorer browsing - I feel we might need to use the win32 api for this to work.

## Commit Summary
- Add imgui library, v1.82 tag
- Integrate imgui with CMake
- Integrate imgui init with application
- Add `onUpdateGUI` to layer base class
- Call `onUpdateGUI` on each layer, each frame, in between Imgui::beginFrame and Imgui::endFrame
- Add an example for using this API to the "levelEditorLayer"


Burke-Daniel (approved):
GTM, send it

liamjwarren (approved):
Lets go

PR #59, opened 2021-04-05
Window Resize

colton-smith:
## Brief
Using WindowResizedEvent-regsistered callbacks, perform appropriate operations when the window is resized.

## Summary
- Add callback in `Application.cpp` to update the viewport size when the window is resized
- Add callback in `Camera.cpp` to update the projection matrix when the window is resized
- Add callback in Window to update the window width and height when the window is resized
- In `PongLayer.cpp`, remove the call to disable cursor


Burke-Daniel (approved):
Nice to see that this was easy to accomplish using events, GTM

liamjwarren (approved):
GTM

PR #60, opened 2021-04-05
Text Rendering from Bitmap font

Burke-Daniel:
This PR adds very simple and unextensible text rendering using a hex display style font. I've added a scoreboard to the pong example, and I haven't added all the characters in the bitmap to the `characterLocations` map, but this can easily be done when more symbols are necessary.


liamjwarren (approved):
Looks good to me, I built and ran on Mac and it displays correctly.

colton-smith (approved):
GTM - Interesting that bitmap derives from spritesheet, makes sense!

PR #62, opened 2021-04-06
Player Functionality - Place Various Items 

colton-smith:
## Brief 
This PR adds some new features to the rpg demo. The player can now cycle through his available items using scroll wheel (up or down - cyclical baby - iterators are wondrous) and place the active item with mouse 1. 

The API allowed this relatively easily which is a good sign! Even the fact that I could do this in a couple hours (irregardless of the spaghetti) is a positive. 

Checkout and check-it-out!

### Engine-specific Changes
- Block input polling from returning true when imgui is consuming mouse and keyboard inputs. This should be added to each event type as well (in the OG callbacks), but for now I have been checking this in my own callbacks to keep things local to my demo and leave engine wide decisions for later


Sandbox/RPG/Toolbar.h
```diff
@@ -0,0 +1,42 @@
...
+	void setColor(const glm::vec4& color);
+
+private:
+	glm::vec2 m_position;
+	glm::vec2 m_size;
+	Aquarius::Texture* m_activeItemIcon;
+	Aquarius::QuadTexCoords* m_activeItemTexCoords;
+	ToolbarAlignment m_alignment;
+	glm::vec4 m_color = { 0, 0, 0, 1 };
+};
```
liamjwarren: Line here

liamjwarren (approved):
This is awesome man! Can show off the sound now too with a sound effect when an item is dropped. Just one line at EOF.

Sandbox/RPG/Level.h
```diff
@@ -57,7 +61,7 @@ class Level
 	int m_tileHeight;
 	int m_tileWidth;
 
-	// TODO Could be an array
 	std::vector<std::vector<Tile>> m_map;
+	std::vector<Item*> m_items;
```
Burke-Daniel, line 65: It's possible for these to be unique pointers right? I'm good with leaving it the way it is though because it really makes no difference

Sandbox/RPG/LevelEditorLayer.cpp
```diff
@@ -94,74 +129,63 @@ void LevelEditorLayer::onUpdate(Aquarius::timeDelta_t time)
...
 	AQ_TRACE("FPS: %v", 1000.0 / time);
 }
 
 void LevelEditorLayer::onUpdateGUI(Aquarius::timeDelta_t time)
 {
+
 	if (m_MenuOpen)
 	{
-		Aquarius::Gui::ShowDemoWindow(&m_MenuOpen);
+		ImGui::Begin("RPG Demo", &m_MenuOpen);   // Pass a pointer to our bool variable (the window will have a closing button that will clear the bool when clicked)
```
Burke-Daniel, line 162: Ahh I was wondering where the close button came from

Sandbox/RPG/LevelEditorLayer.h
```diff
@@ -16,6 +17,10 @@ class LevelEditorLayer : public Aquarius::Layer
 
 private:
 	bool m_MenuOpen = true;
+	float m_MoveSpeed = 0.2;
+	float m_toolbarColor[4] = { 0, 0, 0, 1 };
```
Burke-Daniel, line 21: Love how easy it is to achieve runtime configuration through ImGUI by doing stuff like this

Sandbox/RPG/Player.h
```diff
@@ -177,6 +200,95 @@ class Player
 		m_playerPreviousState = m_playerState;
 	}
 
+	void addItem(Item* item)
+	{
+		// If the bucket for this item type exists
```
Burke-Daniel: this is for when the bucket doesn't exist right? If so then this must be a late night comment hahaha

Sandbox/RPG/Player.h
```diff
@@ -177,6 +200,95 @@ class Player
...
+	int getItemCount(const std::string& item)
+	{
+		if (m_inventory.find(item) == m_inventory.end()) 
+		{
+			return 0;
+		}	
+		else 
+		{
+			return m_inventory[item].size();
+		}
```
Burke-Daniel, line 226: Where's the ternary operator?!?! jkjk don't do that, this is more readable

Sandbox/RPG/Player.h
```diff
@@ -177,6 +200,95 @@ class Player
...
+
+				if (item == *m_activeItemIt)
+				{
+					m_activeItemIt = m_itemTypes.erase(m_activeItemIt);
+					if (m_activeItemIt == m_itemTypes.end())
+					{
+						m_activeItemIt = m_itemTypes.begin();
+					}
+				}
+			}
```
Burke-Daniel, line 253: If this runs when an item of a certain type runs out, do you think it would be more performant to maintain the empty vector, and just check if the particular inventory vector is empty when iterating through the whole inventory to skip over items that have an empty vector?

This is just me thinking out loud about performance, mainly I am just curious if it more performant to maintain empty item vectors and skip over them upon iteration, as opposed to destructing when empty, and then constructing again if an item of that type is picked up in the hypothetical RPG for example

Sandbox/RPG/Player.h
```diff
@@ -187,6 +299,61 @@ class Player
...
+			{
+				if (m_activeItemIt == m_itemTypes.begin())
+				{
+					m_activeItemIt = m_itemTypes.end() - 1;
+				}
+				else
+				{
+					m_activeItemIt--;
+				}
+			}
```
Burke-Daniel, line 335: Very nice very nice

Sandbox/RPG/Toolbar.cpp
```diff
@@ -0,0 +1,65 @@
...
+			break;
+		case ToolbarAlignment::BottomRight:
+			m_position = { screenW - m_size.x, screenH - m_size.y };
+			break;
+		case ToolbarAlignment::TopCenter:
+			m_position = { (screenW / 2) - (m_size.x / 2), 0 };
+			break;
+		case ToolbarAlignment::BottomCenter:
+			m_position = { (screenW / 2) - (m_size.x / 2), screenH - m_size.y };
+			break;
```
Burke-Daniel, line 58: Do you potentially want to make this configurable through ImGUI?

Burke-Daniel (approved):
This is super cool!!! I was playing around with it and everything seems to work as expected. Most of my comments are just me thinking out loud, but one thing I'm wondering about is if we wanna make toolbar alignment configurable through ImGUI? Other than that I'd say GTM

Sandbox/RPG/Toolbar.cpp
colton-smith: Yes - but not now

PR #64, opened 2021-04-06
Sound

liamjwarren:
Brought in AudioEngine manually to allow for necessary modifications to interact with OpenAL-Soft. This branch allows for short sound effects to be played such as the paddle noise heard on the pong application. Currently, you can only load in .wav files due to the loading mechanism in AudioEngine. This is definitely not the cleanest code but it works.


liamjwarren:
Also, let me know if you want me to remove the tests from the AudioEngine library, that will decrease the number of files in the diff.

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -0,0 +1,76 @@
...
+            Initialize();
+        }
+
+        void Device::Initialize()
+        {
+            m_Device = alcOpenDevice(nullptr);
+
+            if (!m_Device)
+            {
+                AQ_CORE_ERROR("Failed to open a sound device");
```
Burke-Daniel, line 21: If the device doesn't open, you probably wanna return from the function here, if not I feel like something undesirable might happen below when you call `alcCreateContext`

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -0,0 +1,76 @@
...
+                AQ_CORE_INFO("A sound device was successfully opened");
+            }
+
+            m_Context = alcCreateContext(m_Device, nullptr);
+
+            if (!m_Context) {
+                AQ_CORE_ERROR("Failed to create the sound context");
+            } else {
+                AQ_CORE_INFO("Sound context created successfully");
+            }
```
Burke-Daniel: Put the brackets on new line instead of same line for consistency

Aquarius/Src/Aquarius/Sound/Devices.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <AL/alc.h>
+#include <memory>
+
+
+namespace Aquarius {
+
+    namespace Sound {
```
Burke-Daniel, line 9: Should probably put empty line below this for consistency

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+            if (!alcMakeContextCurrent(m_Context))
+            {
+                AQ_CORE_ERROR("Failed to make the context current");
+            }
+        }
+
+        Device * Device::get()
+        {
+            static Device* device = new Device();
```
Burke-Daniel: This doesn't get deleted anywhere, I'm wondering if this could cause a memory leak

Aquarius/Src/Aquarius/Sound/SoundBuffer.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <string>
+
+
+namespace Aquarius {
+
+    namespace Sound {
+
+        SoundBuffer* SoundBuffer::get()
+        {
+            static SoundBuffer* sound_buffer = new SoundBuffer();
```
Burke-Daniel: Same comment as Device class

Aquarius/Src/Aquarius/Sound/SoundBuffer.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+            uint32_t format;
+
+            AudioFile<float> sound_file;
+
+            bool loaded = sound_file.load(file_path);
+
+            if(!loaded)
+            {
+                AQ_CORE_ERROR("Unable to open audio file %v, due to %v", file_path);
```
Burke-Daniel, line 41: May want to return from function here as I wonder if the sound file doesn't load correctly will you be able to call `getNumChannels` and other functions on the AudioFile object below here if it is not loaded

Sandbox/PongLayer.h
```diff
@@ -6,6 +6,8 @@
 #include "Pong/Paddle.h"
 #include "Pong/PaddleController.h"
 
+#include <future>
```
Burke-Daniel: This was from me screwing around and is not necessary so I can either remove it or you can, just let me know

Burke-Daniel (approved):
Couple little style things, I ignored some other small style things but nbd, also If you wanna review the part I added then that would be good. Definitely want @colton-smith to take a look at this since we both have code in here now hahaha

Sandbox/PongLayer.h
liamjwarren: I have removed.

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -12,13 +12,19 @@ namespace Aquarius {
             Initialize();
         }
 
+        uniquePtr<Device> Device::Create()
```
Burke-Daniel: So this class is not meant to be a singleton right? Just making sure cause the way this is now you could theoretically have multiple audio devices, unlike before

Aquarius/Src/Aquarius/Sound/SoundBuffer.cpp
```diff
@@ -74,18 +75,6 @@ namespace Aquarius {
...
         }
 
-        SoundBuffer::SoundBuffer()
-        {
-            m_EffectBuffer.clear();
-        }
-
-        SoundBuffer::~SoundBuffer()
-        {
-            alDeleteBuffers(m_EffectBuffer.size(), m_EffectBuffer.data());
```
Burke-Daniel: I think you still might want this to exist, because if it is anything like OpenGL then this will not be handled by the unique pointer destructor

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -40,32 +50,21 @@ namespace Aquarius {
...
-        Device * Device::get()
-        {
-            static Device* device = new Device();
-            return device;
-        }
-
-        Device::~Device()
-        {
-            Deallocate();
-        }
```
Burke-Daniel: You might want this to still exist because idt the unique pointer destructor will handle this

Burke-Daniel (changes requested):
A couple things I am wondering about WRT memory management, I'm also gonna rebase pong on top of this and get a slider added in for the sound volume

Aquarius/Src/Aquarius/Sound/Devices.cpp
liamjwarren: I think there probably should only be one audio device, I will switch it back.

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+            if (!alcMakeContextCurrent(m_Context))
+            {
+                AQ_CORE_ERROR("Failed to make the context current");
+            }
+        }
+
+        Device * Device::get()
+        {
+            static Device* device = new Device();
```
Burke-Daniel: I am not getting any more errors now when this object gets destroyed, but I don't think this will ever get deleted (there is no call to delete on this Device allocated on the heap even though it will only ever get allocated once). It might be better to store this static Device* in the class, and then in get you can check if it is nullptr, and if it is then you can create the Device, and then in the destructor you can delete it, if that makes sense

Aquarius/Src/Aquarius/Sound/Devices.cpp
```diff
@@ -45,15 +45,27 @@ namespace Aquarius {
...
-            static Device* device = new Device();
+            if (device == nullptr)
+            {
+                device = new Device();
+            }
+
+            else
+            {
+                device = nullptr;
+            }
```
Burke-Daniel: Don't need this, this would actually cause a memory leak as you are removing the only reference to a heap object without deleting it first. The above if statement is sufficient, it just creates the device if it doesn't already exist, and if not it just returns it

Aquarius/Src/Aquarius/Core/Window.cpp
```diff
@@ -60,6 +60,9 @@ namespace Aquarius {
         m_Context = RenderingContext::Create(m_Window);
         m_Context->Initialize();
 
+        // Initialize Sound
+        m_SoundDevice->get();
```
Burke-Daniel, line 64: I was wondering how this works but I realized that get is a static method, seems like this class doesn't need to actually have a pointer to a sound device because it doesn't call any methods on it, so this works fine.

Burke-Daniel (approved):
Just one comment but all GTM on my end after Colton takes a look

colton-smith (approved):
GTM - Some style changes and things we can look into after, for the demo GTM.

PR #65, opened 2021-04-06
Pong Enhancements

Burke-Daniel:
This brings some changes to pong to make it more customizable

Main features:

- Configuration menu where paddle and ball speed, paddle length, paddle controller type, scoreboard color can all be changed at runtime
- Scoreboard which properly counts score
- Text which appears at center of screen indicating a point has been scored
- Handling resizing of window properly

There are quite a few static variables within pongLayer.cpp, I could just move some of these into the class but that is just the way I did it for now, definitely some places that could be refactored, but the main thing for now is that it works.

Let me know if you guys see any more opportunities for features that could be added.


liamjwarren (approved):
Very nice, GTM

colton-smith (approved):
GTM

